PR #3, opened 2021-09-16
Updated system test thresholds

sei-nwinski:


sei-amellinger (approved):
Those values should work for me.

PR #12, opened 2021-10-18
Added support for building shadow model experiments from an attack co…

sei-nwinski:
…nfig; added transforms to dataset config


sei-amellinger:
We now have a new branch for the shadow work.

PR #21, opened 2021-11-03
Added support for disabling dt2 validations loss computation.

sei-amellinger:
To test this add the "disable_val_loss" flag to the detectron2 training stanza.

``` python
    "detectron2": {
        "disable_val_loss": true,
```


sei-nwinski (approved):

sei-mchurilla (approved):
Tested ensured that default behavior is to disable

PR #24, opened 2021-11-08
JSON to HJSON

sei-jwhelland:
Switching filesystem config loader from STL json to hjson. The main utility of this extension is improving the user workflow for writing configs. Users can now add comments to their JSON files, which is very handy when it comes to documentation and ad-hoc experimentation -- the user no longer needs to create backup copies of their configs when they want to make changes, they can just comment out lines directly. 

From a backend perspective, the change is seamless -- HJSON is a superset of JSON and can thus parse any plain JSON file.


juneberry/filesystem.py
```diff
@@ -126,9 +148,9 @@ def load_file(path: str):
     """
     # TODO: Add check for '.yaml' and use pyyaml
     if Path(path).exists():
-        if Path(path).suffix == '.json':
+        if Path(path).suffix.lower() == {'.json', '.hjson'}:
```
sei-amellinger, line 151: Should this be "in" instead of ==?

juneberry/filesystem.py
```diff
@@ -115,7 +116,28 @@ def save_json(data, json_path, *, indent: int = 4) -> None:
     :return: None
     """
     with open(json_path, "w") as json_file:
-        json.dump(data, json_file, indent=indent, default=json_cleaner, sort_keys=True)
+        hjson.dump(data, json_file, indent=indent, default=json_cleaner, sort_keys=True)
+
+
+def load(file: BinaryIO, data_type='json') -> dict:
```
sei-amellinger, line 122: What is the API goal for this long term?  To also load yaml files?  Should we just make this the load chokepoint?

juneberry/filesystem.py
sei-jwhelland: Yes, was too hasty resolving merge conflicts during the meeting lol.

juneberry/filesystem.py
sei-jwhelland: Eventually loading yaml files might be useful. I was trying to match the json API of having separate `load` and `loads` functions. It makes more sense to me to go with these functions over `load_file`.

juneberry/filesystem.py
```diff
@@ -115,7 +116,7 @@ def save_json(data, json_path, *, indent: int = 4) -> None:
     :return: None
     """
     with open(json_path, "w") as json_file:
-        json.dump(data, json_file, indent=indent, default=json_cleaner, sort_keys=True)
+        hjson.dump(data, json_file, indent=indent, default=json_cleaner, sort_keys=True)
```
sei-amellinger, line 119: This causes an error in the detectron2 trainer. For detectron2 we create a new json files one for training and one for validation which dt2 reads in directly; we don't pass it an in-memory data structure.  So, in this case we really need to save stuff as pure json because so that dt2 can read it in.

So, for now we probably want an api like:
-save_json which saves json
-save_hjson which saves hjson

Then, we can have someone implement "save_file" which looks at the extension and does the right thing.

sei-jwhelland: This should be addressed by [the most recent commit](https://github.com/cmu-sei/juneberry/pull/24/commits/c5e80bed13ee07473ea1836f041097e7de8448d5).

sei-amellinger (approved):
Looks good.  I'd like to get the new containers working, but I don't want to block the pipeline.

PR #28, opened 2021-11-10
Jb 249 better category handling

sei-vturri:


juneberry/config/coco_utils.py
```diff
@@ -255,7 +255,7 @@ def convert_predictions_to_annotations(predictions: list) -> list:
     return annos
 
 
-def convert_predictions_to_coco(coco_data: dict, predictions: dict) -> dict:
+def convert_predictions_to_coco(coco_data: dict, predictions: dict, category_mapping: dict = False) -> dict:
```
sei-amellinger, line 258: What are the implications of `dict = False` instead of `dict = None`?

juneberry/data.py
```diff
@@ -993,29 +1003,160 @@ def get_label_mapping(model_manager: ModelManager = None, model_config: ModelCon
...
+    if 'categories' in coco_data:
+        return coco_data["categories"]
+    else:
+        return []
+
+
+def categories_in_model_config(model_config_path) -> list:
+    model_config_data = ModelConfig.load(model_config_path)
+    preprocessor_dict = model_config_data.preprocessors
+    object_relabel_dict = next((item for item in preprocessor_dict if item["fqcn"] == "juneberry.transforms"
```
sei-amellinger, line 1047: Can you explain what is going on here?

juneberry/data.py
```diff
@@ -993,29 +1003,160 @@ def get_label_mapping(model_manager: ModelManager = None, model_config: ModelCon
...
+        if category_dict:
+            category_list = []
+            for k, v in category_dict.items():
+                category_list.append({'id': int(k), 'name': v})
+            return category_list
+
+    return []
+
+
+def check_mappings(category_mapping: list, eval_manifest_path: Path, show_source: bool, source: str, source_path: str):
```
sei-amellinger, line 1061: These should have function headers or use _notation to indicate internal methods.

juneberry/mmdetection/evaluator.py
```diff
@@ -228,8 +228,16 @@ def format_evaluation(self) -> None:
         # NOTE: It is currently unclear what the score is
         # [batch] x [num_classes] x [ bbox(l,t,r,b] + score ]
 
+        # Grab category mapping
+        eval_manifest_path = self.model_manager.get_eval_manifest_path(self.eval_dataset_config.file_path)
+        category_mapping = jb_data.get_category_mapping(eval_manifest_path=eval_manifest_path,
```
sei-amellinger, line 233: BTW: Mapping makes me think dict not list.

juneberry/data.py
```diff
@@ -993,29 +1003,160 @@ def get_label_mapping(model_manager: ModelManager = None, model_config: ModelCon
...
+                                         source_info=eval_config.file_path,
+                                         label_dict=label_dict)
+
+
+def categories_in_dataset_config(config_path, data_root) -> list:
+    dataset_config = DatasetConfig.load(config_path)
+    coco_path = dataset_config.image_data.sources[0]['directory']
+
+    # TODO: use future coco prodict
+    with open(data_root / coco_path) as json_file:
```
sei-amellinger, line 1036: Please remember filesystem.load_file(). :)

juneberry/data.py
```diff
@@ -993,29 +1003,160 @@ def get_label_mapping(model_manager: ModelManager = None, model_config: ModelCon
...
+        with open(training_manifest_path) as json_file:
+            train_manifest_data = json.load(json_file)
+        if 'categories' in train_manifest_data:
+            category_mapping = train_manifest_data['categories']
+            if category_mapping:
+                return check_mappings(category_mapping, eval_manifest_path, show_source, "train manifest", str(training_manifest_path))
+
+        # Check the validation manifest
+        validation_manifest_path = model_manager.get_validation_data_manifest_path()
+        with open(validation_manifest_path) as json_file:
```
sei-amellinger, line 1122: filesystem.load_file()

juneberry/data.py
```diff
@@ -993,29 +1003,160 @@ def get_label_mapping(model_manager: ModelManager = None, model_config: ModelCon
...
+    :param eval_config: The DatasetConfig object for model evaluation.
+    :param data_root: The path to the data root of source images.
+    :param show_source: Set to True to return the source from which the label names were extracted.
+    :return: A dictionary containing the category mapping.
+    """
+
+    if model_manager:
+        # Check the training manifest
+        training_manifest_path = model_manager.get_training_data_manifest_path()
+        with open(training_manifest_path) as json_file:
```
sei-amellinger, line 1113: filesystem.load_file()...

test/test_data.py
```diff
@@ -771,3 +773,81 @@ def test_get_label_mapping():
 
     func_labels = jb_data.get_label_mapping(model_config=model_config, train_config=train_config, show_source=False)
     TestCase().assertDictEqual(func_labels, test_labels)
+
+
+def safe_open_w(path):
+    """Open "path" for writing, creating any parent directories as needed."""
+    os.makedirs(os.path.dirname(path), exist_ok=True)
```
sei-amellinger, line 780: Prefer pathlib over os.path

juneberry/config/coco_utils.py
```diff
@@ -256,26 +235,24 @@ def convert_predictions_to_annotations(predictions: list) -> list:
...
-
-    # Make a series of annotations from the predictions
-    for pred in predictions:
-        anno = copy.copy(pred)
-        anno['area'] = pred['bbox'][2] * pred['bbox'][3]
-        anno['id'] = obj_id
-        anno['iscrowd'] = 0
-        obj_id += 1
-        annos.append(anno)
+def convert_predictions_to_coco(coco_data: dict, predictions: list, category_list: List = None) -> dict:
```
sei-amellinger, line 238: The more I think about it, I think we wan to use the term detections as much as possible when in the OD space.  The box is a detection and the class is technically the prediction.

sei-amellinger (commented):
A smattering of different thoughts and tweaks.  We should probably consolidate some terms.

juneberry/data.py
sei-vturri: This iterates through the preprocessor dictionary until it finds (if such a thing exists) the fqcn key corresponding to the object relabel. Then it extracts the category info from the kwargs associated with that entry.

juneberry/config/coco_utils.py
sei-vturri: In the branch I just pushed, I swapped out the word "mapping" for "list' to improve clarity. Do you think I should actually use some other language, like detection?

sei-vturri:
Open issue -- Getting this warning message when running run_all_tests.py: "Category ids in annotations are not in [1, #categories]! We'll apply a mapping for you."

juneberry/config/coco_utils.py
sei-amellinger: Let's post the comment in the group chat. But I suspect "convert_predictions_to_coco" is better as "convert_predictions_to_coco" and we should start changing those names. By start I mean all new occurrences should use the new language, change anything close, but don't spend too much time on it.

juneberry/config/coco_utils.py
```diff
@@ -361,20 +340,15 @@ def save_predictions_as_anno(data_root: Path, dataset_config: str, predict_file:
...
-    # HACK HACK HACK - Assume just one annotations file!
-    # TODO: Add a dataset config/coco_utils for providing a unified config
-    # TODO: Remove mixedCase call eventually
-    if 'imageData' in dataset:
-        coco_path = dataset['imageData']['sources'][0]['directory']
-    else:
-        coco_path = dataset['image_data']['sources'][0]['directory']
-    coco_data = jbfs.load_file(data_root / coco_path)
+    with open(data_root / coco_path) as json_file:
+        coco_data = json.load(json_file)
```
sei-amellinger, line 347: JBFS load file?

juneberry/data.py
sei-amellinger: The ObjectRelabel thing is an example of how to relabel but people can write their own. So, after any preprocessing they might have an entirely new set of categories.

Meaning, I don't think we can safely know if they rewrote or not, so that is why the check the manifest. Perhaps the best we could do is warn if they have any model config preprocessors?

juneberry/data.py
```diff
@@ -993,29 +1011,171 @@ def get_label_mapping(model_manager: ModelManager = None, model_config: ModelCon
...
+    :param data_root: The path to the data root of source images.
+    :param show_source: Set to True to return the source from which the label names were extracted.
+    :return: A dictionary containing the category mapping.
+    """
+    model_config_path = None
+    if model_manager:
+        # Check the training manifest
+        training_manifest_path = model_manager.get_training_data_manifest_path()
+        train_manifest_data = jbfs.load_file(str(training_manifest_path))
+        if 'categories' in train_manifest_data:
```
sei-amellinger, line 1131: Okay, I am going to geek out for a second on style.  You could use the "get" call to reduce a few lines and make it more succinct. So something like:
```
category_list = train_manifest_data.get('categories', None)
if category_list:
```
Not change you need to make now, just a style thing to think about in the future. :)

juneberry/data.py
sei-vturri: Okay, should I not pull the category list from the model config then? And instead, check if there is an ObjectRelabel and warn the users?

sei-amellinger: I think that is probably the best we can hope for. Thoughts?

sei-amellinger (approved):

PR #29, opened 2021-11-11
Shadow v4

sei-amellinger:


bin/jb_run_experiment
```diff
@@ -196,6 +180,16 @@ def main():
             show_command=show_command,
         )
 
+    # Alternative method to generate rules.json using an attack config.
+    attack_mgr = jbfs.AttackManager(args.experimentName)
+    attack_path = attack_mgr.get_experiment_attack_file()
+    if not rules_path.exists() and attack_path.exists():
+        logger.info(f"rules.json in experiment {args.experimentName} not found.")
+        logger.info(f"Generating experiment rules.json file for {args.experimentName} using the attack config.")
+        jbworkflow.run_command(dependencies=[],
```
sei-amellinger, line 189: The attack script should move into bin, be in the setup.py and use the jb_xxx naming scheme.

juneberry/tensorflow/trainer.py
```diff
@@ -125,6 +130,12 @@ def setup(self) -> None:
         np.random.seed(self.model_config.seed)
         tf.random.set_seed(self.model_config.seed)
 
+        # Configure tensorflow to progressively allocate GPU memory rather than fully
```
sei-amellinger, line 133: We don't need this twice do we?

experiments/shadowExperiment/attack.json
```diff
@@ -0,0 +1,59 @@
+{
+    "format_version": "0.1.0",
+    "models": {
+        "shadow": "model_tests/lfw_baselines/leinonet-5-d/",
```
sei-amellinger, line 4: I don't have this model.

juneberry/schemas/attack_schema.json
```diff
@@ -0,0 +1,47 @@
...
+                "disjoint_args": { "type": "array", "items": { "type": "object" } },
+                "private_disjoint_args": { "type": "object" },
+                "private_subset_args": { "type": "object" },
+                "query_watermarks": { "$ref": "#/definitions/plugin" },
+                "superset_args": { "type": "array", "items": { "type": "object" } },
+                "training_watermarks": { "$ref": "#/definitions/plugin" }
+            }
+        }
+    },
+    "required": [ ]
```
sei-amellinger, line 46: Is nothing actually required?

juneberry/filesystem.py
```diff
@@ -194,40 +194,275 @@ def load_file(path: str):
...
+    def get_shadow_model_dir(self, model_num: int, disjoint=False):
+        """
+        :param model_num: Integer that identifies a particular shadow model subdirectory.
+        :param disjoint: Boolean that controls whether to return the 'superset' (default) or
+        'disjoint' version of the desired property.
+        :return: The path to a particular shadow model subdirectory within either the
+        'superset' (default) or 'disjoint' shadow model subdirectory.
+        """
+        model_dir = self.get_shadow_subdir(disjoint=disjoint) / str(model_num)
+        if not model_dir.exists():
```
sei-amellinger, line 403: Path calls should probably not have side effects like this. For example, cleaning code is going to create a directory as a side effect of checking to see if the file is there.

juneberry/schemas/attack_schema.json
sei-nwinski: I added the required properties in 5b9088d.

juneberry/filesystem.py
sei-nwinski: Addressed in e61391f.

bin/jb_run_experiment
sei-nwinski: It looks like this was addressed in b775c34.

sei-nwinski (approved):

PR #30, opened 2021-11-12
check for empty annos in Metrics init and evaluators

sei-wshaw:


juneberry/config/coco_utils.py
```diff
@@ -471,7 +471,11 @@ def generate_bbox_images(coco_json: Path, lab, dest_dir: str = None, sample_limi
     logger.info(f"Added bounding boxes to {img_count} {img_str}.")
     logger.info(f"Drew {box_total} {box_str} across all images.")
 
-
+def count_annotations(coco_json: Path) -> int:
```
sei-amellinger, line 474: Can we switch this to taking strings?  In general we've been switching to use Path inside every function but the APIs are string based.

juneberry/detectron2/evaluator.py
```diff
@@ -192,10 +192,22 @@ def evaluate_data(self) -> None:
...
 
-        # Populate metrics
-        m = metrics.Metrics.create_with_filesystem_managers(self.model_manager, self.eval_dir_mgr)
-        self.output.results.metrics.bbox = m.as_dict()
-        self.output.results.metrics.bbox_per_class = m.mAP_per_class
+        anno_file = Path(self.eval_dir_mgr.get_manifest_path())
+        num_annotations = coco_utils.count_annotations(anno_file)
+
+        # Populate metrics if there are annotations
+        if num_annotations > 0:
```
sei-amellinger, line 199: PyCharm is telling me that there are 11 identical lines to the MMD evaluator.  Lines 260+.  Can we refactor those out?

juneberry/config/coco_utils.py
sei-wshaw: Yes, that makes sense, I'll change it.

juneberry/detectron2/evaluator.py
sei-wshaw: Sure. Question though; those lines are duplicated between the detectron2 and mmd evaluators. At first I thought, just move those lines to the superclass, but when I look at it, the superclass evaluator doesn't have much in the way of business logic. It seemed weird to put them in there. If it's ok to move them there I will, but wanted your thoughts.

sei-amellinger: The current base class isn't really a good idea.

We actually don't have a good organization right now.  The evaluator is a blend of classification and object detection code and really should be split more clearly.  So, putting more OD specific stuff in "EvaluatorBase" is only going to muddle it.  We need to move to splitting classification and OD into separate base classes.

The training results are are also too big and need to be split, so this doesn't make it any easier.

There is an a utils under juneberry.evaluation.  Would it fit there?

sei-wshaw: I moved it into utils, you can have a look. There's a little problem running in the mmd evaluator. When I get back from today's meeting I'll fix it.

juneberry/evaluation/utils.py
```diff
@@ -59,9 +63,11 @@ def get_histogram(dataset_dicts, classes):
     histogram = np.zeros((num_classes,), dtype=np.int)
     for entry in dataset_dicts:
         annos = entry["annotations"]
-        classes = np.asarray([x["category_id"] for x in annos if not x.get("iscrowd", 0)], dtype=np.int)
+        classes = np.asarray(
+            [x["category_id"] for x in annos if not x.get("iscrowd", 0)], dtype=np.int)
         if len(classes):
-            assert classes.min() >= 0, f"Got an invalid category_id={classes.min()}"
+            assert classes.min(
```
sei-amellinger, line 69: What editor are you using?  BTW: We are using 120 chars and it looks like it is doing some funky wrapping.

sei-wshaw: Sorry I didn't respond, was using VS Code but didn't intend to format that. I fixed it.

juneberry/detectron2/evaluator.py
sei-wshaw: There was no problem in mmd eval, so never mind that comment.

sei-amellinger (approved):

PR #31, opened 2021-11-19
Potential fix to the off-by-one image ID issue

sei-nwinski:


sei-mchurilla (approved):
Code looks correct. Evaluated new bounding box drawing and it is on the correct image.

PR #34, opened 2021-12-03
Generate sample eval manifest categories

sei-vturri:


sei-amellinger:
Now when I run it complains there is no training_manifest available.

path = 'models/text_detect/dt2/ut/train/training_manifest.json'

So it looks like we have the problem elsewhere.  So to test this I made sure I had nothing in my text_detect/dt2/ut directory except the config (and known_ut_results.json) and ran `pytest test/test_data.py`.  It looks like some other files will need to be symthesized.

sei-amellinger (approved):

PR #36, opened 2021-12-08
Pinned torch on cpudev and revved bug fix number.

sei-amellinger:


sei-nwinski (approved):
Built the cpudev container without issue; system test results looked good.

PR #37, opened 2021-12-09
Create workspace, run container start in workspace

sei-wshaw:


docker/enter_juneberry_container
```diff
@@ -32,12 +32,15 @@
...
 #   workspace <- This is an optional external workspace directory, see docker/README.md for details
 
-# Get the project root from the first argument.
-if [ ${#} -ne 1 ]; then
-  echo "This script requires one argument: a path to the 'project' root."
-fi
-# PROJ=${HOME}/proj
+
+# Show usage and exit if we don't get at least one command line argument.
+[ $# -eq 0 ] && { echo "Usage: $0 <project> [workspace]"; exit 1; }
```
sei-amellinger, line 37: Maybe something like if > 2 args then whine?

docker/enter_juneberry_container
```diff
@@ -32,12 +32,15 @@
...
+
+# Set PROJ and WS (project and workspace) to the first and second cmdline args.
 PROJ=${1}
+WS=${2}
+
+echo "Using project ${PROJ}..."
```
sei-amellinger, line 43: workspace

scripts/setup_workspace.py
```diff
@@ -0,0 +1,105 @@
...
+
+    src_dir = workspace_dir / "src"
+    src_dir.mkdir()
+
+def copy_container_start(project_dir, workspace_dir):
+    container_start = Path(project_dir, "juneberry/docker/container_start.sh")
+    shutil.copy(container_start, workspace_dir)
+    
+def main():
+    parser = argparse.ArgumentParser()
```
sei-amellinger, line 67: Consider adding a description and help to each field.

docker/enter_juneberry_container
sei-wshaw: Ok, is the current whining sufficient (meaning change -eq 0 to -gt 2 and that's all)

scripts/setup_workspace.py
```diff
@@ -0,0 +1,105 @@
...
+    args = parser.parse_args()
+
+    project_dir = Path(args.project_dir)
+    workspace_dir = Path(args.workspace)
+    
+    # If the workspace directory exists and is a directory...
+    if workspace_dir.exists() and workspace_dir.is_dir():
+
+        # If the workspace directory is not empty...
+        if os.listdir(workspace_dir):
```
sei-amellinger, line 79: There is a pretty good chance the workspace is not empty.  They may have checked out one from something like bitbucket.  However the workspace might have a juneberry.ini or a container_start because we usually don't check those in.  So, you should detect to see if they don't exist and then add them.  

So in short this is the use case:
- I have a working juneberry setup
- I check out an existing workspace (e.g. cornelius-workspace) from bitbucket
- I run it specifying that workspace

Have you tested this workflow?

sei-wshaw: I was assuming if you give me a non-empty directory and it doesn't contain a juneberry.ini, it's not a workspace and exit. So no. Should we just accept any non-empty directory as a possible workspace then, or do a sanity check for things like "models", etc.

docker/enter_juneberry_container
sei-amellinger: I think the prose is sufficient.

scripts/setup_workspace.py
sei-amellinger: The juneberry.ini is machine specific so it is part of the .gitignore.  So yeah, we need to add one.  

I think it would be nice to add models and experiments directories if they don't have them.

sei-wshaw: Ok that's fine. So if we see a non-empty directory, we'll try to make it a compliant Juneberry workspace (without being destructive of course).

sei-amellinger: Even if it is empty we should set it up?

sei-wshaw:
To test:

- Starting as ``juneberry/docker/enter_juneberry_container `pwd` `` starts the container with /juneberry as the workspace. (default behavior)
- Starting as ``juneberry/docker/enter_juneberry_container `pwd` <workspace_dir>`` starts the container with `workspace_dir` as the workspace.
   - If `workspace_dir` doesn't exist, one will be created.
   - `workspace_dir` will contain `juneberry.ini`, `data_sets`, `models`, and `src`. If any of these don't exist, they will be created.
- `juneberry/docker/container_start.sh` is a sample container start script that can copied to `/juneberry` (for the default case), or to `workspace_dir`. It should run automatically.

docker/cpudev.Dockerfile
```diff
@@ -82,3 +82,5 @@ RUN echo "PS1='${debian_chroot:+($debian_chroot)}\u@\h+CPUDev:\w\$ '" >> /root/.
     echo "    echo './container_start.sh NOT found.'"  >> /root/.bashrc; \
     echo "fi" >> /root/.bashrc
 
+ENV PATH="${PATH}:${PROJ}/juneberry/bin"
```
sei-amellinger, line 85: Should we get rid of this?

docker/cudadev.Dockerfile
```diff
@@ -83,3 +83,5 @@ RUN echo "PS1='${debian_chroot:+($debian_chroot)}\u@\h+CudaDev:\w\$ '" >> /root/
     echo "    echo './container_start.sh NOT found.'" >> /root/.bashrc; \
     echo "fi" >> /root/.bashrc
 
+ENV PATH="${PATH}:${PROJ}/juneberry/bin"
```
sei-amellinger, line 86: And this?

docker/cpudev.Dockerfile
sei-wshaw: yes

docker/cudadev.Dockerfile
sei-wshaw: yes

sei-amellinger (approved):

PR #40, opened 2021-12-14
Output trained models in onnx format

sei-nwinski:
Here are some of the features gained by this work:

1. jb_train can now produce model.onnx files in addition to typical platform specific formats. See the new jb_train args for usage details.
2. You can evaluate any ONNX models produced by the Juneberry PyTorch or TensorFlow trainers. There have been some additions to the model config specification that describe how to accomplish this. The raw predictions data produced by these ONNX evaluations is expected to match the raw predictions data from the "standard" evaluations by at least 3 (and often more) decimal places.
3. There's a new model directory named `FasterRCNN-10` which contains a README with some prep steps and the bare minimum model config that you'll need in order to conduct an evaluation of a Faster R-CNN ONNX model from the ONNX model zoo. `juneberry.onnx.faster_rcnn` will contain the classes you need to perform this eval.


juneberry/onnx/default.py
```diff
@@ -0,0 +1,118 @@
...
+        """
+        When called, this method uses the attributes of the evaluator to format the raw evaluation data. The
+        result of the process is the evaluator.output attribute will contain JSON-friendly data, which will
+        then be written to a file.
+        :param evaluator: The Evaluator object managing the evaluation.
+        :return: Nothing.
+        """
+
+        # Add the predicted labels for each image to the output.
+        labels = [item[1] for item in evaluator.eval_name_targets]
```
sei-amellinger, line 54: Pycharm is telling me that lines 54-85 are duplicates of juneberry.pytorch.evaluation.default.py lines 71-102.

sei-nwinski: Addressed in 849ecdc.

juneberry/onnx/default.py
```diff
@@ -0,0 +1,87 @@
...
+
+        # Perform the common eval output processing steps for a classifier.
+        jb_eval_utils.prepare_classification_eval_output(evaluator)
+
+        # Calculate the hash of the model that was used to conduct the evaluation.
+        evaluated_model_hash = jbfs.generate_file_hash(evaluator.model_manager.get_onnx_model_path())
+
+        # If Juneberry was used to train the model, retrieve the hash from the training output file
+        # and verify the hash matches the hash of the model used to evaluate the data.
+        training_output_file_path = evaluator.model_manager.get_training_out_file()
```
sei-amellinger, line 61: This is very similar to the code in the juneberry/pytorch/evaluation/default.py  Can you refactor this into juneberry/evaluation/utils.py?  I didn't see any call like this in the tensorflow stuff.  I wonder if we should add it there too?

juneberry/onnx/onnx_model_zoo/faster_rcnn.py
```diff
@@ -0,0 +1,292 @@
...
+#
+# DM21-0884
+#
+# ======================================================================================================================
+
+import json
+import logging
+import numpy as np
+from PIL import Image
+from tqdm import tqdm
```
sei-amellinger, line 29: Does this come from onnx or should we install this with our setup?

sei-nwinski: We'll want tqdm in our setup. I'm trying to use it in more places, where it makes sense. This is what displays the progress bars.

juneberry/onnx/default.py
sei-nwinski: Addressed in abc7eda.

juneberry/onnx/onnx_model_zoo/faster_rcnn.py
sei-amellinger: Okay. I added an issue for it.

sei-amellinger (approved):

PR #45, opened 2022-01-10
Feature/jb 362 update docker images

sei-wshaw:


sei-wshaw:
- I had to pin the opencv-python version to match the version of opencv-python-headless in cudadev. If these don't match, unit tests fail.
- run_system_test on cpudev has an error: 2022-01-07 23:39:13,005 ERROR Evaluation mAP for text_detect/dt2/ut of 3.8299997443599036e-05 did NOT exceed min mAP: 4e-05. See evaluation output file for details.
Is this difference small enough not to matter?

sei-amellinger:
> * run_system_test on cpudev has an error: 2022-01-07 23:39:13,005 ERROR Evaluation mAP for text_detect/dt2/ut of 3.8299997443599036e-05 did NOT exceed min mAP: 4e-05. See evaluation output file for details.
>   Is this difference small enough not to matter?

I think it is okay for now.  We are in the process of trying to pin down where the differences come from.

sei-nwinski:
> * I had to pin the opencv-python version to match the version of opencv-python-headless in cudadev. If these don't match, unit tests fail.
> 
>     * run_system_test on cpudev has an error: 2022-01-07 23:39:13,005 ERROR Evaluation mAP for text_detect/dt2/ut of 3.8299997443599036e-05 did NOT exceed min mAP: 4e-05. See evaluation output file for details.
>       Is this difference small enough not to matter?

Line 103 in scripts/run_system_test.py is where the mAP: 4e-05 threshold is set. Historically this was just set to the lowest observed value for mAP since we're not deterministic here yet and there's some variation in the metric. The assumption was that so long as you exceeded the minimum observed threshold, you should "pass" the test.

sei-amellinger (approved):

PR #47, opened 2022-01-19
Added in options for specifying machine class

sei-vturri:
Specify machine-class via (in order of preference):
- command-line (-c, --machineClass)
- environment variable (JUNEBERRY_MACHINE_CLASS)
- juneberry.ini (machine_class)

### DIFF
--- a/juneberry/config_loader.py
+++ b/juneberry/config_loader.py
@@ -115,7 +115,6 @@ def setup_lab(overrides: dict, section_name: str = None):
     }
     lab_args = {}
 
-
     # Walk each input value and create arguments for the lab object
     errors = 0
     for k in KEYS_TYPES.keys():
--- a/juneberry/config_loader.py
+++ b/juneberry/config_loader.py
@@ -115,6 +115,7 @@ def setup_lab(overrides: dict, section_name: str = None):
     }
     lab_args = {}
 
+
     # Walk each input value and create arguments for the lab object
     errors = 0
     for k in KEYS_TYPES.keys():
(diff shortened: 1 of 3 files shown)


sei-amellinger:
If I specify an environment variable and something in the ini file, the ini file wins. I think we want the env variable to get priority. Thoughts?

juneberry/config_loader.py
```diff
@@ -173,6 +153,12 @@ def setup_lab(overrides, section_name=None):
             lab_args[required] = None
             errors += 1
 
+    # Check environment variable if machine_class not found in command line or ini
+    if 'machine_class' not in lab_args:
+        machine_class = os.environ.get('JUNEBERRY_MACHINE_CLASS')
+        if machine_class is not None:
+            lab_args['machine_class'] = str(machine_class)
+
     # Now return a lab args object initialized to these values
```
sei-nwinski, line 162: I was running a little test where I would print out the lab_args at this line under various conditions. Here's my setup:
`machine_class = a.b.c.d.e` in my juneberry.ini
$JUNEBERRY_MACHINE_CLASS set to `a.b.c.d`

I then ran a jb_train command with the added args `-c a.b.c`. This caused my lab_args print statement to have a machine_class of a.b.c, which is what I expected.

Next, I removed the `-c` arg from the same jb_train command and ran it again. This time my print statements showed a machine_class of a.b.c.d.e, which I didn't expect based on the priority order. Shouldn't it have been a.b.c.d?

sei-amellinger: Violet, I'm glad you pointed that out. Do we actually have workspace and dataroot in env variables? From what I can see we have a path to an ini in an environment variable (which is lowest in priority) but not the actual variables.

sei-amellinger (approved):

juneberry/config_loader.py
sei-vturri: I don't think we do either. Looks to me like we only use these env variables (+ the new machine class variable): JUNEBERRY_CUDA_MEMORY_SUMMARY_PERIOD, CUDA_VISIBLE_DEVICES, and JB_REMOVE_OLD_LOG. Are these env variables in the documentation?
Also, I could add in checks for workspace and dataroot env variables in this PR if that sounds good. Would we want the same priority order as for machine_class?

sei-nwinski (approved):

PR #48, opened 2022-01-19
Bugfix/jb 368 logging issues

sei-nwinski:
Lots of logging changes here. The starting point is probably in `juneberry/scripting.py`. You'll see that a new arg has been added for setting a directory where the log file will be saved, with a default of the current working directory. 

Once that was in place, the next area to look would be the new function in `scripting.py` which was built to take advantage of that new arg. Basically any script can use this function to set up the root juneberry logger and dump a log file to the desired log directory. Scripts where logging was already set up the way we wanted weren't updated to use this new function, but most of the other scripts were.

The last major change to point out would be a new `logs` subdirectory that gets created inside of an experiments directory. The overall experiment logs should start showing up in this `logs` directory now, and any script logs (for things like summary reports, etc.) will also be placed here if they're associated with the experiment.

### DIFF
--- a/bin/jb_experiment_to_rules
+++ b/bin/jb_experiment_to_rules
@@ -74,13 +74,12 @@ class RuleMaker:
             inputs = [model_mgr.get_model_config(), model_config.training_dataset_config_path]
             outputs = model_mgr.get_training_output_list()
             command = [JB_TRAIN_COMMAND, model_mgr.model_name]
-            doc = f"{JB_TRAIN_COMMAND} {model_mgr.model_name}"
 
             # Add the "--onnx" option to jb_train if an ONNX model was requested.
             if model.onnx:
                 command.append("--onnx")
-                doc += " --onnx"
 
+            doc = " ".join([str(x) for x in command])
             clean_extras = model_mgr.get_training_clean_extras_list()
             requirements = []
 
@@ -92,7 +91,7 @@ class RuleMaker:
             inputs = [model_mgr.get_model_config(), model_config.training_dataset_config_path]
             outputs = model_mgr.get_dry_run_output_list()
             command = [JB_TRAIN_COMMAND, model_mgr.model_name, '--dryrun']
-            doc = f"{JB_TRAIN_COMMAND} {model_mgr.model_name} --dryrun"
+            doc = " ".join([str(x) for x in command])
             clean_extras = model_mgr.get_dry_run_clean_extras_list()
             requirements = []
 
@@ -128,7 +127,7 @@ class RuleMaker:
                         command.append('--useTrainSplit')
                     if test.use_val_split:
                         command.append('--useValSplit')
-                    doc = f"{JB_EVALUATE_COMMAND} {model_mgr.model_name}  {test.dataset_path}"
+                    doc = " ".join([str(x) for x in command])
                     clean_extras = model_mgr.get_predictions_clean_extras_list()
                     requirements = [model['train_rule_id']]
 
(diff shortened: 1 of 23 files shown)


juneberry/scripting.py
```diff
@@ -130,6 +132,28 @@ def setup_workspace(args, *, log_file, log_prefix="", add_data_root=True, model_
...
+    the root juneberry logger cannot be set up by other means.
+    :param args: The args from parseargs.
+    :param script_name: A string indicating which script the log messages are from. Used to build
+    the name of the log file.
+    :return: Nothing.
+
+    """
+    # Build the location of the log using the desired log directory (from args) and the name of
+    # the script calling this function.
+    log_file = Path(args.logDir) / f"log_{script_name}.txt"
```
sei-amellinger, line 147: If this is really the script name and we want it to be a little easier to use, we could use
```
if script_name is None:
    script_name=Path(sys.argv[0]).stem
```
As a way so each script doesn't have to enter it, and if someone renamed the script during debugging, it would automatically reflect that in the log. A small feature with the renaming thing, but it means that if we do rename scripts we have one less place to do so.

Thoughts?

sei-nwinski: I like it, but I'm going to want to verify that argv[0] comes back as expected in the context of an experiment/rules file. I wasn't a huge fan of how I implemented the creation of the log file names because of those hardcoded script names, so I'm definitely willing the experiment a little more in this area.

bin/jb_process_dataset
```diff
@@ -354,19 +354,19 @@ def main():
...
+            logger.error(f"As a safety measure we will not overwrite an output directory. Please delete the output "
+                         f"directory and try again. Exiting.")
             sys.exit(-1)
 
     logger.info(f"Making output directory {str(out_dir_path)}")
     out_dir_path.mkdir()
 
     # Follow the normal procedure for setting up logs and workspace
     log_prefix = ""
-    log_file = str(out_dir_path / "process_log.txt")
```
sei-amellinger, line 366: Before, the log file was kept with the newly created dataset, sort of as a record of what was created and why. Now, (by default) it is stored locally (cwd). Sure a person could explicitly specify logDir now, but can we have the default (in this case) be the output dir?

juneberry/scripting.py
sei-amellinger: Absolutely! The subprocess (and doit stuff) should behave similarly in that they sub-python shell should get the right argv, but that stuff can always be a little tricky.

While your updating scripting and looking at this you can also log the full path the executable?  So an `absolute()` from the sys.argv[0]?  I always though it would be nice to log precisely which script we are running.

bin/jb_experiment_to_rules
```diff
@@ -188,7 +189,7 @@ class RuleMaker:
         outputs = [Path(output_dir, "pr_curve.png"), Path(output_dir, "pc_curve.png"),
                    Path(output_dir, "rc_curve.png"), Path(output_dir, "eval_metrics.csv")]
 
-        doc = f"{JB_PLOT_PR_COMMAND} --outputDir {output_dir}"
+        doc = f"{JB_PLOT_PR_COMMAND} --outputDir {output_dir} --logDir {log_dir}"
```
sei-amellinger, line 192: I should have written this as:
```
doc = " ".join(command)

juneberry/scripting.py
sei-nwinski: Addressed in 0cf4078. argv[0] testing looked good in experiments/pydoit, and I added another logging message showing the absolute path of the script.

bin/jb_process_dataset
sei-nwinski: Addressed in 0cf4078. I went with a solution that could potentially conflict with a situation where the user specifies that the log directory should be the current directory, but I included a warning if the change is made.

bin/jb_experiment_to_rules
sei-nwinski: Addressed in 0cf4078, but you'll see it a little differently than what you thought. It was a little cranky because not everything in command was always a string (I experienced ints and Paths). Instead of identifying which fields could potentially come in as non-strings, I figured it was just safer to add a list comprehension inside the join.

sei-amellinger (approved):

PR #50, opened 2022-01-24
method to compute tp, fp, fn and corresponding test

sei-wshaw:
This adds a call to the Metrics API to get TP, FP, FN.


sei-mchurilla (approved):
Verified this works on  real files / data

PR #52, opened 2022-01-27
JB-372 extend DT2 config path

sei-amellinger:
Added ability to specify full model config via module.
Added ability to add supplemental configs.
Updated docs.

To test:
- Run a dry run and save off the platform config.
- Copy config_with_sample_platform.json to config.json
- Run again and compare two platform configs.  The real difference should be the seed.


juneberry/detectron2/trainer.py
```diff
@@ -527,6 +522,36 @@ def extract_dt2_log_content(self, file: Path):
...
+        # Looks through the supplements property and merge in those files
+        if self.model_config.detectron2 is not None:
+            if self.model_config.detectron2.supplements is not None:
+                for path in self.model_config.detectron2.supplements:
+                    tmp_path = Path(path)
+                    if tmp_path.exists():
+                        logger.info(f"Loading config supplements from {path}")
+                        cfg.merge_from_file(path)
+                    else:
+                        logger.error(f"Failed to find supplement {path}. Exiting")
```
sei-nwinski, line 552: This seems a little harsh, but maybe it was requested. I would think if a supplement was requested but not found, you would just try to move on without the supplement. Log a warning for sure, but not terminate the attempt.

sei-nwinski: I looked at the PyTorch classifier trainer, DT2 trainer, and MMD trainer. I thought there were more, but the only examples I found where we move on after not doing something the config asks for are: ignoring model transforms if they are requested in DT2, and ignoring a requested validation split if the dataset is a Torchvision dataset.

So while not all misconfigurations, there is precedent. Thinking about it some more though, I've had a change of heart about terminating here. I think it's better to force the user to review their supplements and correct any errors versus training with a skipped supplement that was a "must have" for what the user was trying to accomplish.

PR #60, opened 2022-02-04
Exit enter_juneberry_container if setup_workspace returns non-zero

sei-wshaw:


sei-nwinski:
In `setup_workspace.py`, I think the key name in L44 needs to be "workspace_root" and the key_name in L46 needs to be "tensorboard_root".

sei-wshaw:
Ok, so we should fix up docker/juneberry.ini as well? Reading through the code in config_loader.py line 108ff, it seemed like "workspace" and "tensorboard" are the correct keys, but there might be something going on there I don't grok.

sei-nwinski (approved):

PR #64, opened 2022-02-08
Feature/jb 369 add machine class to lab

sei-vturri:
- Created a machine_specs prodict
- Added a machine_specs load to Lab
- Wrote test cases for the machine_specs loader
- Removed all "hints" from the repo


juneberry/detectron2/trainer.py
```diff
@@ -190,10 +190,6 @@ def setup(self) -> None:
         if self.num_gpus == 0:
             cfg.MODEL.DEVICE = "cpu"
 
-        # AOM: I have no idea of the impact of this
```
sei-amellinger, line 193: If we remove this, where are we setting the cfg.DATALOADER.NUM_WORKERS?

juneberry/config/machine_specs.py
```diff
@@ -0,0 +1,181 @@
...
+
+                    else:
+                        for prop in MachineSpecs.properties:
+                            if prop in config_data[machine][key]:
+                                specs_data[prop] = config_data[machine][key][prop]
+
+        return specs_data
+
+    @staticmethod
+    def check_workspace_config(config_data: dict):
```
sei-amellinger, line 103: Check meaning invalidate?

juneberry/config/machine_specs.py
```diff
@@ -0,0 +1,181 @@
...
+        :param machine_class: The name of the class of machine.
+        :param model_name: The name of the model.
+        :return: Loaded, validated, and constructed object.
+        """
+        # Load the raw file.
+        logger.info(f"Loading MACHINE CONFIG from {data_path}")
+        config_data = jbfs.load_file(data_path)
+        specs_data = {}
+
+        # Checks config
```
sei-amellinger, line 158: Check meaning validate?

juneberry/config/machine_specs.py
```diff
@@ -0,0 +1,181 @@
...
+        MachineSpecs.check_workspace_config(config_data)
+
+        # Check default:default
```
sei-amellinger, line 161: Check meaning load from?

juneberry/config/machine_specs.py
```diff
@@ -0,0 +1,181 @@
...
+            for key in config_data[machine].keys():
+                if re.match(key, model):
+                    if "include" in config_data[machine][key]:
+                        include_lst = config_data[machine][key]["include"].split(':')
+                        machine = include_lst[0]
+                        model = include_lst[1]
+                        specs_data = MachineSpecs.update_properties(machine, model, config_data, specs_data)
+
+                    else:
+                        for prop in MachineSpecs.properties:
```
sei-amellinger, line 96: I'm not sure we want to do limit the fields.  Let's chat about this when you have time.

docs/specs/model_configuration_specification.md
```diff
@@ -327,10 +324,6 @@ Linux style version of the **format** of the file. Not the version of
 the data, but the version of the semantics of the fields of this file. 
 The current version: 0.2.0
 
-## hints
-**Optional:** A section for different values that may improve that performance or general computation. These should 
-not impact the output in any substantial way and may be ignored by the system based on other requirements.
-
 ### num_workers
```
sei-nwinski, line 327: This explanation for num_workers would also need to be removed since the parent (hints) no longer exists.

juneberry/config/machine_specs.py
```diff
@@ -0,0 +1,181 @@
...
+# Copyright notice for non-US Government use and distribution.
+#
+# This Software includes and/or makes use of Third-Party Software subject to its own license.
+#
+# DM21-0884
+#
+# ======================================================================================================================
+
+import logging
+from prodict import List, Prodict
```
sei-nwinski, line 26: List wasn't used anywhere.

juneberry/config/machine_specs.py
```diff
@@ -0,0 +1,181 @@
...
+        if error_count > 0:
+            logger.error(f"Found {error_count} errors in machine specs. EXITING.")
+            sys.exit(-1)
+
+    @staticmethod
+    def construct(data: dict):
+        """
+        Load, validate, and construct a machine specs object.
+        :param data: The data to use to construct the object.
+        :param file_path: Optional path to a file that may have been loaded. Used for logging.
```
sei-nwinski, line 66: You can remove this file_path line.

juneberry/config/machine_specs.py
```diff
@@ -0,0 +1,181 @@
...
+    def _finish_init(self) -> None:
+        """
+        Validate machine specs fields
+        :return: None
+        """
+        error_count = 0
+
+        # Check that the number of gpus doesn't exceed the number of cuda visible devices
+        if self.num_gpus and os.environ.get('CUDA_VISIBLE_DEVICES'):
+            if self.num_gpus > int(os.environ.get('CUDA_VISIBLE_DEVICES')):
```
sei-nwinski, line 53: I don't think this will work. Can't you have something like `CUDA_VISIBLE_DEVICES=0,1`? Won't that create a problem here?

juneberry/detectron2/trainer.py
sei-nwinski: The Lab should already exist by the time you get to this point, so you can probably set this with self.lab.num_workers and just not have it triggered by the presence of "hints" in the model_config.

sei-nwinski:
A few more places to remove "hints" from:
 - known_warnings.md has a section named "Number of workers in ModelConfig". With hints gone, this warning shouldn't appear again.
 - models/text_detect/dt2/ut/config_with_sample_platform.json has a "hints" section.

sei-nwinski:
Would it make sense to add a unit test that verifies a Lab gets its num_gpus, num_workers, and machine class set correctly?

bin/jb_train
```diff
@@ -192,10 +182,6 @@ def main():
         # Allocate the GPUs the user asked for, getting the current number if None are required.
         trainer.num_gpus = trainer.check_gpu_availability(args.num_gpus)
 
-        if max_gpus is not None and trainer.num_gpus > max_gpus:
```
sei-amellinger, line 195: Where is the gpu number passed into the trainer?

juneberry/config/machine_specs.py
```diff
@@ -0,0 +1,183 @@
...
+            for key in config_data[machine].keys():
+                if re.match(key, model):
+                    if "include" in config_data[machine][key]:
+                        include_lst = config_data[machine][key]["include"].split(':')
+                        machine = include_lst[0]
+                        model = include_lst[1]
+                        specs_data = MachineSpecs.update_properties(machine, model, config_data, specs_data)
+
+                    else:
+                        for prop in config_data[machine][key].keys():
```
sei-amellinger, line 98: Would `specs_data.update(config_data[machine][key])` work?

juneberry/config/machine_specs.py
```diff
@@ -0,0 +1,183 @@
...
+import juneberry.filesystem as jbfs
+
+logger = logging.getLogger(__name__)
+
+
+class MachineSpecs(Prodict):
+    """
+    A class to validate and manage the machine specifications pulled from the machine config file.
+    """
+    properties = ["num_workers", "num_gpus"]
```
sei-amellinger, line 40: Can we get rid of this now?

juneberry/config/machine_specs.py
```diff
@@ -0,0 +1,183 @@
...
+        :param config_data: The machine config.
+        :param specs_data: The dictionary of specs for execution.
+        :return: The machine specification object.
+        """
+
+        if machine in config_data:
+            for key in config_data[machine].keys():
+                if re.match(key, model):
+                    if "include" in config_data[machine][key]:
+                        include_lst = config_data[machine][key]["include"].split(':')
```
sei-amellinger, line 92: Let's add a little error checking. Maybe?
```
try:
  (machine, model) = config_data[machine][key]['include'].split(':')
except:
  print(f"The include {config_data[machine][key]['include']} doesn't match format of 'machine:model'")
```
Or some such (I didn't syntax check this suggestion.)

juneberry/config/machine_specs.py
```diff
@@ -0,0 +1,183 @@
...
+        default_error_count = 0
+
+        if "default" not in config_data:
+            default_error_count += 1
+        else:
+            if "default" not in config_data["default"]:
+                default_error_count += 1
+
+        # Report missing default:default section
+        if default_error_count > 0:
```
sei-amellinger, line 116: I don't think we need to require default:default.  We should just fall back to our system defaults.

juneberry/config/machine_specs.py
```diff
@@ -0,0 +1,183 @@
...
+            logger.error(f"Config missing default:default section.")
+
+        # Check for valid include
+        include_error_count = 0
+
+        for machine_key in config_data.keys():
+            for model_key in config_data[machine_key]:
+                if "include" in config_data[machine_key][model_key]:
+                    include_lst = config_data[machine_key][model_key]["include"].split(":")
```
sei-amellinger, line 125: Similar to above we should check the format for two fields, has colon, etc.  Probably make a function like "parse_include" or something and then re-use everywhere.

juneberry/config/machine_specs.py
```diff
@@ -0,0 +1,183 @@
...
+        :return: The machine specification object.
+        """
+
+        if machine in config_data:
+            for key in config_data[machine].keys():
+                if re.match(key, model):
+                    if "include" in config_data[machine][key]:
+                        include_lst = config_data[machine][key]["include"].split(':')
+                        machine = include_lst[0]
+                        model = include_lst[1]
```
sei-amellinger, line 94: Is this model the model that is the function arg?  I am confused as to why you are changing it here.

juneberry/lab.py
```diff
@@ -96,6 +97,17 @@ def load_dataset_config(self, config_path, ws_key='default') -> DatasetConfig:
...
+    def load_machine_specs(self, model_name, ws_key='default') -> MachineSpecs:
+        """
+        Finds the relevant execution specs based on machine and model class.
+        :param model_name: The model name.
+        :param ws_key: The workspace to use to load.
+        :return: A MachineSpecs object containing execution specifications.
+        """
+        machine_class = self.machine_class
+        machine_config_path = Path(self.workspace(ws_key)) / 'config.json'
+        return MachineSpecs.load(data_path=str(machine_config_path), machine_class=machine_class, model_name=model_name)
```
sei-nwinski, line 109: You'll get an error here when you try to run the system test and you don't have a MACHINE CONFIG defined in a file. It tries to load {WORKSPACE_ROOT}/config.json which fails to load because it does not exist.

Also this was pretty confusing when I first saw it because the config.json made me think I need to look for a missing model config. So I guess that's +1 reason why we should name these configs something else to differentiate them from model configs.

sei-amellinger: We definitely should not require a workspace config. 

Regarding path names, I am still resistant to redundant and exhaustive naming. We need to adopt a clear and simple philosophy and stick with it.
Meaning, if explicit naming is the way to go, do you prpose to rename `experiments/<foo>/config.json` to `experiments/foo/experiment_config.json` and `models/foo/bar/config.json` to `models/foo/bar/model_config.json`?

The rules should be simple and minimal. We need to train the everyone that location is as important the name and leverage that, or the name is what matters.  Location will be important (the model path *is* the model name), etc. then I prefer we continue to emphasize our documentation and understanding of layout.

sei-nwinski: If I see a config.json inside an experiments path, I assume experiment config. Same thing for the model config example. Take a look at this error message:

`<<DRY_RUN>> 2022-02-17 13:10:48,837 ERROR Failed to load /juneberry/config.json. The file could not be found. EXITING.
`
Following our path pattern, would you assume the file is a Juneberry config? There's more context around that message that leads you down the path of "it's a MACHINE CONFIG":

`<<DRY_RUN>> 2022-02-17 13:10:48,835 INFO Loading MACHINE CONFIG from /juneberry/config.json`

but I'm trying to say that path + filename isn't really enough to let someone make an educated guess about what type of config it is.

sei-amellinger: Technically it is a workspace config, so the messages should read loading the machine class from the workspace config.

You say "...path + filename isn't really enough to let someone make an educated guess about what type of config it is" so you are saying we should change model and experiment configs too?

bin/jb_train
sei-vturri: Currently the gpu number is passed as a command line argument:
parser.add_argument('-n', '--num-gpus', type=int, default=None, help='The number of GPUs. By default use all. Set to 0 for CPU.')

sei-vturri: From the command line num_gpus argument

sei-amellinger (approved):

PR #67, opened 2022-02-28
Feature/jb 387 remove ini

sei-amellinger:


docker/enter_juneberry_container
```diff
@@ -2,146 +2,151 @@
...
+# JUNEBERRY_CONTAINER       - The name of the container to run.
+# JUNEBERRY_GPUS            - The GPUS for the container: "" for none, "all" to set to all, ids for specific ones.
+# JUNEBERRY_MACHINE_CLASS   - The machine class
+
+# === Lab Layout ===
+LAB=${PWD}
+
+# If they specified a lab directory then use it as a basis for the lab structure.
+if [ ! -z ${JUNEBERRY_LAB} ]; then
+    LAB=${PWD}
```
sei-nwinski, line 46: My bash isn't great, but I think there might be something off here. 

Shouldn't we have `LAB=${JUNEBERRY_LAB}` somewhere? Like inside this `if`?

sei-amellinger: Good catch!

Fixed.

docs/building_docker.md
```diff
@@ -39,20 +39,18 @@ When the containers start up they will look for a script called "container_start
...
 * /tensorboard - Mount point for tensorboard output
 * /root/.cache/torch/hub - Mounted for model caches for PyTorch and MMDetection
 * /root/.torch - Mounted for model caches for Detectron2
 
-As a convenience in the "home" directory (/root) is a juneberry.ini file that specifies the /juneberry
-and /datasets directories as the workspace and dataroot respectively.
-Thus, when logging into a clean container, __if no juneberry.ini exists in /juneberry__ then juneberry will
-find the juneberry.ini in the /root directory and automatically find the juneberry and datasets directories.
+The containers set the dataroot and tensorboard environment variables according.  We use the 
+current working directory as the workspace.
```
sei-nwinski, line 53: I'm struggling with these two lines. I feel like they need a bit of clarification. Would the following rewrite make sense:

The containers set the dataroot and tensorboard variables automatically. The current working directory will be chosen as the workspace, unless specified otherwise.

docs/overview.md
```diff
@@ -1,11 +1,38 @@
...
+<project>
+    juneberry/ - The cloned repo for sample projects
+    dataroot/ - Where all the data assets are stored.
+    tensorboard/ - OPTIONAL - Where tensorboard outputs are stored
+    cache/ - A place where we store platform cached data for pytorch, tensorflow, etc.
+    *workspace-directories* - e.g. juneberry-example-workspace
+```
+
+This structure is preferred but not required. Sometimes, for larger deployments, it is important
+to store files in paths that could be different storage devices. Later, we'll discuss how to create your own.
```
sei-nwinski, line 21: "Later, we'll discuss how to create your own."

This line wasn't clear to me on two points. I couldn't figure out what was going to be created (lab structure? workspace?) and also where the creation of this thing was being talked about (later? elsewhere?).

docs/overview.md
```diff
@@ -22,24 +49,14 @@ of the configuration is done via files.
...
-Normally, it is much more convenient to specify these (and other settings) once via a `juneberry.ini` file, described
-below in "Configuration via juneberry.ini". If you wish to use a `juneberry.ini` file instead of providing directories
-via command line switches, you are expected to create your own `juneberry.ini` file in an appropriate location. Refer to
-the "Search Order" section of "Configuration via juneberry.ini" for the relevant details.
-
-# WORKSPACE_ROOT Directory
-
-The workspace path can be set via the `-w` command line switch or more generally via the `workspace_root` field in your
-`juneberry.ini` file.
+The [Getting Started](getting_started.md) explains how to specify the workspace.
```
sei-nwinski, line 59: This line is redundant given the previous sentence, right? Or is there another point we're trying to link to?

docs/building_docker.md
sei-amellinger: Sounds good! I suspect you have other tweaks in the works, otherwise I can make the change.

sei-nwinski: Yeah, I'll get take care of this.

docs/overview.md
sei-amellinger: Yes, this should be written.  The "your own" is confusing but is meant towards "file structure." And I don't think we actually do that clearly later. I had some stuff before, but tore it out for expedience. Basically we have a set of environment variables/command line switches accepted by jb tools that can be set for a custom environment. 

At this stage basically "creating your own environment" is just "choose whatever layout you like and set all the variables or switches."

So, that section could read something like:

"This structure is preferred but not required. Sometimes, for larger deployments, it is important to store files in paths that could be different storage devices. If a different structure is chosen then the directories must be provided to `enter_juneberry_container` script or the Juneberry tools (e.g. jb_train) via the various configuration variables.  See getting_started.md for details."

docs/overview.md
sei-amellinger: It is redundant.

sei-nwinski (approved):

PR #68, opened 2022-03-08
Fix for experiments where a model is evaluated by multiple datasets

sei-nwinski:
If you have an experiment where you have a trained model, and then you want to evaluate that model multiple times using different datasets, pydoit will complain because the parent eval directory `models/model_name/eval` will always appear in the list of targets (output files) that should get created).

The fix is to simply remove that target from the list of files. That check was redundant anyway, because the first target that gets added to the list is the "root" eval directory `models/model_name/eval/eval_dataset_name`. Since the check for this target will remain, if it exists then this implies that `models/model_name/eval` must also exist.

### DIFF
--- a/bin/jb_run_experiment
+++ b/bin/jb_run_experiment
@@ -92,11 +92,13 @@ def main():
                                                  "a series of outputs into the experimentName directory. NOTE: By "
                                                  "default this script reports what it WOULD do but does not perform "
                                                  "any action. See --commit. NOTE: This script executes other scripts "
-                                                 "in the same directory.")
+                                                 "in the same directory. ARGS NOTE: Any command line args that are not "
+                                                 "recognized as positional or optional args for jb_run_experiment will "
+                                                 "be passed to the pydoit command.")
 
     setup_args(parser)
     jbscripting.setup_args(parser, add_data_root=True)
-    args = parser.parse_args()
+    args, unknown = parser.parse_known_args()
     show_command = args.showcmd
 
     # Set up logging
@@ -172,11 +174,7 @@ def main():
         cmd_args = [bin_dir / "jb_rules_to_pydoit", args.experimentName, workflow]
         if args.processes and args.processes > 0:
             cmd_args.append('--parallel')
-        jbworkflow.run_command(
-            dependencies=[],
-            args=cmd_args,
-            show_command=show_command,
-        )
+        jbworkflow.run_command(dependencies=[], args=cmd_args, show_command=show_command)
 
     # Develop pydoit command
     cmd_args = ["python", "-m", "doit", "-d", str(lab.workspace()), "-f", str(dodo_path.resolve())]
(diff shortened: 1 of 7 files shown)


sei-amellinger (approved):

scripts/run_system_test.py
```diff
@@ -264,7 +264,14 @@ def check_training_metric(model_name, model_mgr, eval_dir_mgr, min_train_metric,
 
     train_outfile = model_mgr.get_training_out_file()
     training_data = TrainingOutput.load(train_outfile)
-    eval_data = EvaluationOutput.load(eval_dir_mgr.get_predictions_path())
+    if model_mgr.model_task is "classification":
```
sei-amellinger, line 267: This should be "==" instead of "is" right?

scripts/run_system_test.py
```diff
@@ -652,12 +659,18 @@ def check_metric(test_set) -> int:
     for model_name, test_name, min_train_metric, min_eval_metric in test_set:
         model_mgr = jbfs.ModelManager(model_name)
         eval_dir_mgr = model_mgr.get_eval_dir_mgr(test_name)
-        prediction_file_path = eval_dir_mgr.get_predictions_path()
+        if model_mgr.model_task is "classification":
```
sei-amellinger, line 662: Same "==" vs "is" question?

scripts/run_system_test.py
sei-nwinski: Fixed in 62d3845

scripts/run_system_test.py
sei-nwinski: Fixed in 62d3845

bin/jb_run_experiment
```diff
@@ -207,6 +205,18 @@ def main():
             cmd_args.append("list")
             cmd_args.append("-s")
 
+    # Save the db file for pydoit to the experiment directory.
+    db_path = str(experiment_manager.get_experiment_db_file())
```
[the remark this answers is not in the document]
sei-nwinski: I thought the goal was to make it so that any db files get saved into the individual experiment directories. Did I misunderstand?

Also, can you clarify what you think would need to be container within the run experiment code?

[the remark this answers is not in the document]
sei-amellinger: Doh! Sorry! Looking back that comment is missing a lot of context.  I am talking about the fact that the path is in the experiment manager. The name of the file is probably only really in scope w.r.t. to the run_experiment and no one else needs to know. I hope that makes a little more sense.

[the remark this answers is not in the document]
sei-nwinski: I guess my thinking on filesystem stuff evolved to the point where anytime Juneberry did something related to file system I just stick in in `filesystem.py`.

I suppose `experiment_manager.experiment_dir_path / .doit.db` here would be equivalent and gets it out of `filesystem.py`. Is that what you'd prefer? I'm fine either way.

sei-amellinger (approved):

PR #73, opened 2022-03-21
added simple transform for replacing the FC layer

sei-nmvanhoudnos:


sei-nwinski (approved):

sei-nwinski:
Full disclosure: I didn't have an example handy to confirm that this transform performs the intended operation, so I'm trusting Nathan/Matt that it does what they need. Format-wise it's fine and it doesn't break anything else in Juneberry (at least that I saw during testing).

PR #76, opened 2022-03-30
Feature/jb 345 pi attack report

sei-nwinski:
This PR adjusts how Juneberry handles Reports. There are no longer individual report scripts for generating Summary, ROC, and PR reports. Instead, jb_report was created to create reports via Juneberry Report classes, located in `juneberry/reporting/`. This PR also introduces a new Report class for property inference attack experiments. 

The work in this PR is strongly linked to PR #2 in the juneberry-example-workspace. The two PRs should be reviewed and merged together.


sei-wshaw:
Error when running the system tests:

FileNotFoundError: [Errno 2] No such file or directory: 'jb_report'
<<LIVE>> 2022-05-05 16:08:34,968 ERROR Command returned error code 3. Command: '['python', '-m', 'doit', '-d', '/juneberry-example-workspace', '-f', '/juneberry-example-workspace/experiments/smokeTests/classify/main_dodo.py', '--db-file', 'experiments/smokeTests/classify/.doit.db']'. EXITING!!
2022-05-05 16:08:34,997 ERROR Returned error code 255. EXITING!!

sei-nwinski:
I fixed the assertion error in the PRReport unit test in 62d3251. I used the wrong default value in the test (should have 0.5 all along, not sure why I thought it was 1.0. Probably confused it with a different field)

bin/jb_report
```diff
@@ -0,0 +1,84 @@
...
+#
+# ======================================================================================================================
+
+import argparse
+import logging
+
+import juneberry.reporting.utils as jb_report_utils
+import juneberry.scripting as jbscripting
+
+logger = logging.getLogger("juneberry.jb_report")
```
sei-wshaw, line 31: Not a huge thing, but are we still explicitly naming the loggers, or just using ```__name__```?

docs/specs/model_configuration_specification.md
```diff
@@ -681,6 +682,11 @@ Any provided accuracy score must take the parameters "y_pred" (array of predicte
 ### accuracy_args
 A dictionary of optional arguments to be passed to the accuracy function.
 
+## reports
+**OPTIONAL** An array of one or more Report Plugins, where each Plugin corresponds to a Juneberry Report
```
sei-amellinger, line 686: Only Juneberry reports?

bin/jb_report
```diff
@@ -0,0 +1,84 @@
...
+    jbscripting.setup_args(parser)
+    args = parser.parse_args()
+
+    # Set up logging.
+    jbscripting.setup_logging_for_script(args)
+
+    report_list = None
+
+    # Fetch the desired reports for the experiment and build each report.
+    if args.experiment:
```
sei-amellinger, line 61: What if multiple different options are selected?  It looks like -f wins and the others are silently ignored.

bin/jb_experiment_to_rules
```diff
@@ -25,30 +25,29 @@
 # Read the experiment config and splat it out to a graph.
 
 import argparse
+import json
 import logging
 from pathlib import Path
+from prodict import List, Prodict
```
sei-amellinger, line 31: BTW: Prodict is not a built-in.

docs/specs/report_configuration_specification.md
```diff
@@ -0,0 +1,468 @@
+Report Configuration Specification
+==========
+
+
+# Introduction
+This document describes the schema for the JSON formatted Juneberry Report configuration file. This 
+schema has a unique behavior when compared to other Juneberry configuration files, in that it may
```
sei-amellinger, line 7: You don't need to change this, but technically it isn't correct. The LabProfile (which has a schema and config model but NOT a spec - bad Andrew) is also used in multiple places.  So, we are seeing an emerging pattern here. So, we have two use cases and probably should have a short chat about any "standards" we want.

docs/specs/report_configuration_specification.md
```diff
@@ -0,0 +1,468 @@
...
+or experiment configs.
+
+# Basic Report Schema
+The following schema applies to standalone ReportConfigs, or any "reports" stanzas that have been 
+inserted into model configs or experiment configs.
+```
+{
+    "reports": [
+        {
+            "description": <Human readable description of this report>,
```
sei-amellinger, line 18: This structure does not seem to be consistently deployed. In config/model.py and config/report.py the reports use the Plugin class, but in experiment config there is a Report class that also has "test" field (which are Pllugins) and in experiment outline there is a different Report class that also has classes. They don't extend Plugin with the new fields but seem to redefine the new fields.  Maybe we should put description into the plugin and then subclass things elsewhere?

Additionally, the schemas don't seem to align.  Unfortunately I never made a plugin item schema separately and that appears to have been propagated as debt. The report schema has a plugin_item but it isn't consistent with the model_schema, and the experiment and experiment_outline also seem to deal with them separately.

Anyway, there seems to be a bit of alignement/refactoring debt here.

I'll leave breadcrumbs elsewhere.

docs/specs/report_configuration_specification.md
```diff
@@ -0,0 +1,468 @@
...
+            "fqcn": <fully qualified name of a class that extends the juneberry.reporting.report base class>,
+            "kwargs": { <OPTIONAL kwargs to be passed (expanded) to __init__ on construction> }
+        }
+    ]
+}
+```
+
+## Report Specific Schemas
+
+Juneberry includes a few built-in Report classes. This section describes the schemas for each
```
sei-amellinger, line 28: These fields don't actually seem to be in the report schema.

docs/specs/report_configuration_specification.md
```diff
@@ -0,0 +1,468 @@
...
+        "accuracy": 0.5
+    }
+}
+```
+
+The previous stanza could then be inserted into the "reports" list of a report config file, a 
+model config file, or an experiment config file.
+
+# Schema Adjustments in an Experiment Config
+When a "reports" stanza exits inside an experiment config, the schema and behavior of a particular
```
sei-amellinger, line 144: ...exists...

docs/specs/report_configuration_specification.md
```diff
@@ -0,0 +1,468 @@
...
+```python
+from juneberry.reporting.report import Report
+
+class ExampleReport(Report):
+    def __init__(self, report_output_dir: str, model_name: str, accuracy: float):
+        super().__init__(output_str=report_output_dir)
+        self.model_name = model_name
+        self.accuracy = accuracy
+        
+    def create_report(self) -> None:
```
sei-amellinger, line 109: So, if the all the variables are set in __init__ and create_report is just parameterless, what is the value of having this as a plugin?

docs/specs/report_configuration_specification.md
```diff
@@ -0,0 +1,468 @@
...
+# Schema Adjustments in an Experiment Config
+When a "reports" stanza exits inside an experiment config, the schema and behavior of a particular 
+Report may vary slight from the schemas described above. This section will describe these differences 
+for each Report Type.
+
+There's a good chance that some filenames required by a particular report's schema may not be 
+known in advance of running the experiment. In these situations, a report stanza may omit certain 
+fields from the experiment config and rely on Juneberry to fill in the correct value(s) for the 
+field once those values are determined. This is especially common for the "curve_sources" field in 
+the ROC and PR Reports. When the pydoit rules for the experiment are generated, Juneberry will fill
```
sei-amellinger, line 152: In general we don't want to be (and really aren't) Doit coupled/dependent. The rules file is workflow agnostic. So this should really just be "When the rules file is generated...".  I suppose we could use "rules.json" or italicize or something.

bin/jb_experiment_to_rules
```diff
@@ -146,139 +145,468 @@ class RuleMaker:
...
-        if report.iou:
-            command.append(f'--iou')
-            command.append(report.iou)
-
-        # Add all tests to the command.
+        outputs = []
+
+        # Use the idx to construct a unique filename for the individual JSON file that will contain
+        # ONLY the information for this report.
+        report_json_name = f"report_{idx}.json"
```
sei-amellinger, line 225: This pattern "report_{idx}.json" shows up throughout this file. Should we refactor this out?

bin/jb_experiment_to_rules
```diff
@@ -146,139 +145,468 @@ class RuleMaker:
...
+        report_json_path = self.experiment_manager.experiment_reports_dir_path / report_json_name
+
+        # The individual JSON file for this report is a product of this process, so add it to the
+        # list of outputs.
+        outputs.append(str(report_json_path))
+
+        # Initialize the contents of the JSON file to match the desired report.
+        report_json_content = report
```
sei-amellinger, line 233: Why assign this? Without a "copy" this is just referencing the passed in values.

bin/jb_experiment_to_rules
```diff
@@ -146,139 +145,468 @@ class RuleMaker:
...
-        # Obtain the output directory as a string.
-        output_dir = str(self.experiment_manager.get_experiment_file(report.output_dir))
-        log_dir = str(self.experiment_manager.get_experiment_log_dir())
-
-        # Accumulate all the tests into one report.
+    def add_reports(self) -> None:
+        # Add Reports when the experiment config contains a "reports" stanza.
+        if self.experiment_config.reports:
+
+            # These properties are used when a Summary Report is not the last Report
```
sei-amellinger, line 152: If we wanted to support multiple summaries, we could just do a two-pass.  The first for not-summary, the second fro summaries.

bin/jb_experiment_to_rules
```diff
@@ -146,139 +145,468 @@ class RuleMaker:
...
+                    self.add_plot_roc(report, idx)
+
+                # Add a properly formatted 'jb_report' rule to generate PR curves
+                # when the report type matches the PLOT_PR class.
+                elif report.fqcn == ReportFQCN.PLOT_PR:
+                    self.add_plot_pr(report, idx)
+
+                # Add a properly formatted 'jb_report' rule to generate a summary
+                # report when the report type matches the SUMMARY class.
+                elif report.fqcn == ReportFQCN.SUMMARY:
```
sei-amellinger, line 184: Perhaps add a safety check that if redo_summary_flag is true, bail.

bin/jb_experiment_to_rules
```diff
@@ -146,139 +145,468 @@ class RuleMaker:
...
+        # The individual JSON file for this report is a product of this process, so add it to the
+        # list of outputs.
+        outputs.append(str(report_json_path))
+
+        # Initialize the contents of the JSON file to match the desired report.
+        report_json_content = report
+
+        # If an output_dir was not provided in the report's kwargs, use the experiment directory.
+        if "output_dir" not in report.kwargs:
+            logger.warning(f"An output_dir was not found in the kwargs of report {idx} in the experiment "
```
sei-amellinger, line 237: Can we suggest a change to something so this warning doesn't happen again? Such as "update {xyz} file with {abc}".

bin/jb_experiment_to_rules
```diff
@@ -146,139 +145,468 @@ class RuleMaker:
...
+
+            # No matter what output_dir the user provided via the config, we want to make sure their desired
+            # directory is inside the current experiment directory. So, resolve both Paths and then check if
+            # the experiment directory matches any of the parent directories for the user's desired output
+            # directory. If the desired output_dir doesn't appear to be inside the experiment dir, then adjust
+            # the output dir to be a sub-dir inside the experiment directory.
+            if self.experiment_manager.experiment_dir_path.resolve() not in output_dir_path.resolve().parents:
+                new_output_dir_str = str(self.experiment_manager.experiment_dir_path / output_dir_path)
+                report_json_content.kwargs.output_dir = new_output_dir_str
+                logger.warning(f"The desired output_dir in report {idx} does not appear to be inside the "
```
sei-amellinger, line 254: As above: Do we have a suggestion for a fix?

bin/jb_experiment_to_rules
```diff
@@ -146,139 +145,468 @@ class RuleMaker:
...
+        for index, report in enumerate(self.experiment_config.reports):
+
+            # When the index values match, that means the loop is looking at the report stanza for the current
+            # summary report. Since no plot files are created by this process, it can be skipped.
+            if index == idx:
+                continue
+
+            # Determine the filename for this report's individual report.json file stored in the experiment
+            # directory. Load the contents of that JSON file.
+            report_file_path = self.experiment_manager.experiment_reports_dir_path / f"report_{index}.json"
```
sei-amellinger, line 480: There is a function that returns the path, should it be used instead of the variable?  Also, another case of the format string being used.

bin/jb_experiment_to_rules
```diff
@@ -146,139 +145,468 @@ class RuleMaker:
...
+            inputs.append(str(report_file_path))
+            with open(report_file_path, 'r') as file:
+                report_file_content = json.load(file)
+
+            # Convert the loaded JSON content to a Prodict (to make retrieving attributes "cleaner").
+            report_prodict = Prodict.from_dict(report_file_content["reports"][0])
```
sei-amellinger, line 486: Should this be "ReportConfig" load(...)

bin/jb_experiment_to_rules
```diff
@@ -146,139 +145,468 @@ class RuleMaker:
...
                 continue
 
+        # If any changes were made to the plot_files, the content for the JSON file should reflect
+        # those changes.
+        report_json_content.kwargs.plot_files = plot_files
+
+        # Create the individual JSON file for this report and add the rule to the workflow.
+        self._finalize_report(report_json_content, report_json_path, inputs, requirements, outputs)
+
+    def add_generic_report(self, report: Report, idx: int) -> None:
```
sei-amellinger, line 517: The generic output has the fqcn and kwargs but could probably use the test_tag, or perhaps the entire test stanza.  Otherwise, it will be hard for generic plugins to know what the test content was without digging through the config.

juneberry/reporting/attack_summary.py
```diff
@@ -0,0 +1,239 @@
...
+        :param output_file: The file where this row will be written.
+        :return: A list of all the accuracy values used in the summary statistics row.
+        """
+        # Construct a model manager for the model and locate the model's training image.
+        model_mgr = jbfs.ModelManager(model_name)
+        train_img_str = model_mgr.get_training_summary_plot()
+
+        # Load the model's training output file and retrieve the accuracy values inside.
+        training_output = TrainingOutput.load(model_mgr.get_training_out_file())
+        # TODO: Right now this retrieves the last item in the list, but we may want the "best" epoch.
```
sei-amellinger, line 106: The last item might not be the one that actually represents the model.  The acceptance checker is currently based on val loss not training accuracy. The acceptance checker is pluggable, so that could change too.

So, we currently have a big gap (and need to log an issue) that we don't actually capture the the epoch end up being the "best". So we currently have no way to actually find out which epoch to report.

But yes, last may not have any bearing on best.

juneberry/reporting/summary.py
```diff
@@ -0,0 +1,289 @@
...
+            dst_training_plot = self.report_files_path / f"{model_name_str}_output.png"
+            copy(orig_training_plot, dst_training_plot)
+
+            # Retrieve the metric value from the metrics data.
+            metric_value = self._retrieve_metric_data(table_data.metrics_stanza)
+
+            # Assemble the table row and add it to the summary table.
+
+            # Accuracy metrics are formatted one way.
+            if self.metric == 'Accuracy' or self.metric == 'Balanced Accuracy':
```
sei-amellinger, line 171: Since these are basically duplicate lines, maybe the if-else just formats the metric value?

docs/specs/experiment_outline_specification.md
```diff
@@ -78,31 +77,36 @@ specification for details.
...
 
 ### description
 **REQUIRED** for any report type. A brief, human-readable description of the report.
 
-### test_tag
-**REQUIRED** if the report type is "plot_roc". This should be a string that matches one of the 
-tags in the 'tests' section. For every model in the experiment, a ROC plot will be produced using 
-the test data from the tag matching the tag provided here.
+### fqcn
+The fully qualified class name (fqcn) of the Report to build, e.g.
```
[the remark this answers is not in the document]
sei-nwinski: You're right, that's an oversight in the schema file. I'm adding a 6th option in the anyOf portion of the schema to accept a generic string in the fqcn.

docs/specs/report_configuration_specification.md
sei-nwinski: I think this is a mistake I made when drafting this introduction. I might have been working/thinking about the schema while working on this. I looked at all of the other introductions in the spec files and none of them use the "schema" language here, so I need to remove it here for consistency.

bin/jb_experiment_to_rules
sei-nwinski: I'll fix this here and adhere to the standard we discussed on 5/9 going forward.

docs/specs/report_configuration_specification.md
sei-nwinski: I think I rewrote the schema files and spec files to address this.

docs/specs/report_configuration_specification.md
sei-nwinski: I think the new Report schema file will address this.

docs/specs/report_configuration_specification.md
sei-nwinski: Fixed.

docs/specs/report_configuration_specification.md
sei-nwinski: I think I understand the issue you're trying to draw attention to, but I don't think I understand what alternative you're proposing. Could you maybe restate what change you're looking for to address this?

docs/specs/report_configuration_specification.md
sei-nwinski: Will change to "When generating the rules.json file for the experiment..."

It didn't feel like italics were necessary in that adjustment.

bin/jb_experiment_to_rules
sei-nwinski: Fixed.

bin/jb_experiment_to_rules
sei-nwinski: Fixed (in multiple spots)

bin/jb_experiment_to_rules
sei-nwinski: Implemented.

bin/jb_experiment_to_rules
sei-nwinski: With the two-pass strategy, the whole "redo summary" process became irrelevant, so this flag no longer exists.

bin/jb_experiment_to_rules
sei-nwinski: Added additional warning messages which suggest fixes for the other warning messages.

bin/jb_experiment_to_rules
sei-nwinski: Added additional warning messages which suggest fixes for the other warning messages.

test/test_reports.py
```diff
@@ -0,0 +1,639 @@
...
+        assert roc_report.legend_font_size == legend_font_size
+        assert roc_report.curve_sources == curve_sources
+
+    def test_output_filename_as_dir(self):
+        """
+        The purpose of this test is to exercise that an ROC Report will set the correct
+        "report_path" when a directory is provided as the value for the "output_filename".
+        """
+        # Establish a directory to test.
+        test_directory = "/test"
```
sei-amellinger, line 443: I think this an all other tests should use the `tmp_path` instead of assuming they have access to the root of the machine.

bin/jb_experiment_to_rules
sei-nwinski: I ended up making a new function to determine which path to use here, which also affected some of the refactoring I did earlier to eliminate the f"report_{idx}.json" bit.

bin/jb_experiment_to_rules
sei-nwinski: Yes. I was able to remove a few lines here in favor of the ReportConfig.load

bin/jb_experiment_to_rules
```diff
@@ -146,139 +145,468 @@ class RuleMaker:
...
+            # summary report. Since no plot files are created by this process, it can be skipped.
+            if index == idx:
+                continue
+
+            # Determine the filename for this report's individual report.json file stored in the experiment
+            # directory. Load the contents of that JSON file.
+            report_file_path = self.experiment_manager.experiment_reports_dir_path / f"report_{index}.json"
+            inputs.append(str(report_file_path))
+            with open(report_file_path, 'r') as file:
+                report_file_content = json.load(file)
```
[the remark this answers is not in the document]
sei-nwinski: I was able to eliminate this load when I addressed an earlier piece of feedback that asked if a ReportConfig.load() made sense here.

juneberry/reporting/summary.py
sei-nwinski: The if-else ended up not being needed. It'll just be a single if that adjusts the formatting when the metric is an accuracy value.

test/test_reports.py
sei-nwinski: I think I was able to all instances where this feedback applied, at least in this file.

bin/jb_experiment_to_rules
sei-nwinski: What if it's a Report type similar to the Summary reports, where the tests don't matter?

bin/jb_report
```diff
@@ -0,0 +1,87 @@
...
+
+
+def setup_args(parser) -> None:
+    """
+    Adds arguments to the parser
+    :param parser: The parser in which to add arguments.
+    """
+
+    # The purpose of the group is to allow only one of either '-e', '-m', or '-f' to be provided.
+    group = parser.add_mutually_exclusive_group()
```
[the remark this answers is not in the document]
sei-nwinski: Good point. It will be included in the next commit.

docs/specs/report_configuration_specification.md
sei-amellinger: This is a longer design conversation that would be easier to do over voice.  We can do it in PL>

bin/jb_experiment_to_rules
sei-amellinger: Then there is nothing to copy. :)

sei-amellinger (approved):

PR #78, opened 2022-04-07
Feature/jb 398 transforms change labels

sei-amellinger:
Transforms can now accept labels as optional arguments and can optionally return (item, label) pairs.

In doing this I cleaned up the way in which optional arguments were passed in for the Staged transform. Previously we specially detected the signature (order was important) but now we just detect all optional arguments (except 'kwargs') past the first. This exposes those fields for general purpose and paves the way for future expansion.

The ugly part was how the return values are processed. We now have a "LabeledTransformManager" subclass that deal with label handling.

### DIFF
--- a/docs/specs/model_configuration_specification.md
+++ b/docs/specs/model_configuration_specification.md
@@ -282,9 +282,11 @@ or testing.
 *Pytorch Transform Format*
 
 Each transform is an instance of a class that has a `__call__` method which accepts an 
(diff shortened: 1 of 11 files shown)


sei-wshaw:
I'm currently getting a test failure:
FAILED test/tf_test/test_data.py::test_dataset - TypeError: <lambda>() got an unexpected keyword argument 'label'

sei-wshaw (approved):

PR #80, opened 2022-04-12
Feature/jb 393 extract workspace

sei-amellinger:
This should be landed at the same time as we land the changes to the example workspace.


scripts/run_system_test.py
```diff
@@ -555,6 +555,7 @@ def check_clean(runner, experiment_name: str, model_names: list, error_summary,
```
sei-nwinski, line 555: Shouldn't this whole file (run_system_test.py) no longer exist in the Juneberry repo? Aren't most of our tests going to fail unless you've brought the workspace along?

docs/getting_started.md
```diff
@@ -171,6 +189,9 @@ source /juneberry/scripts/juneberry_completion.sh
 NOTE: If you use a temporary container (by default `enter_juneberry_container` creates temporary containers)
 you'll need to source the juneberry completion script every time you start the container.
 
+TODO: This section needs a rewrite since the model is now in the example workspace. It seems like we're trying 
+to dance this line of "you don't really need the example workspace" and "yeah you kind of do". We should commit to 
+one side and write that approach.
```
sei-nwinski, line 194: Adding a comment for visibility.

docs/getting_started.md
```diff
@@ -188,6 +209,7 @@ For convenience, users can create a bash script containing the previous commands
 or inside a custom workspace (see below), it will be executed during the container's initialization. The 
 `juneberry/docker` directory contains a sample `container_start.sh` script.
 
+TODO: I think overview.md is probably a better home for this section.
```
sei-nwinski, line 212: Adding a comment for visibility.

docs/overview.md
```diff
@@ -1,6 +1,7 @@
 Workspace and Experiment Overview
 ==========
 
+TODO: Does this section really add anything that hasn't already been explained in getting_started.md#Lab-Layout ?
```
sei-nwinski, line 4: Adding a comment for visibility.

docs/overview.md
```diff
@@ -106,6 +110,8 @@ For more details on the structure of experiment configs, refer to
 [experiment_configuration_specification.md](specs/experiment_configuration_specification.md) in the
 Juneberry documentation directory.
 
+TODO: I think getting_started.md is probably a better home for the next two sections (dataroot and tensorboard) 
+since that's more "lab layout" themed.
```
sei-nwinski, line 114: Adding a comment for visibility.

docs/getting_started.md
sei-amellinger: I did a lot of rework there.  Let me know if this is starting to look better.

docs/getting_started.md
sei-amellinger: I kept it in the getting started for now.  My theory has been getting started is like install and overview is more general layout.

docs/overview.md
sei-amellinger: Maybe not? The idea is that that the getting_started will probably only be read once, where the overview is more of a reference. I am trying to keep getting started as more of a TL;DR with high signal to noise ratio.

scripts/run_system_test.py
sei-nwinski: I think we might be in a bit of a weird spot here, because `run_all_tests.py` relied on `run_system_test.py` which is now in the example workspace, but also needs the unit tests, which continue to live in this repo.

sei-nwinski (approved):

PR #82, opened 2022-04-19
Feature/jb 414 no op training support

sei-amellinger:
A couple simple fixes/ checks for zero-length use cases.

### DIFF
--- a/juneberry/detectron2/trainer.py
+++ b/juneberry/detectron2/trainer.py
@@ -108,6 +108,11 @@ class Detectron2Trainer(Trainer):
         for key in self.results_keys:
             self.output.results.update({key: []})
 
+        # NOTE: This is a custom path for DT2.  We know that DT2 puts the model file in the final
+        # OUTPUT directory, so we rename it out of there.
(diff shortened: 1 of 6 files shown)


juneberry/detectron2/trainer.py
```diff
@@ -108,6 +108,11 @@ def __init__(self, lab: Lab, model_manager: ModelManager, model_config: ModelCon
         for key in self.results_keys:
             self.output.results.update({key: []})
 
+        # NOTE: This is a custom path for DT2.  We know that DT2 puts the model file in the final
+        # OUTPUT directory, so we rename it out of there.
+        self.output_dir = self.model_manager.get_train_scratch_path()
+        self.final_model_path = self.output_dir / 'model_final.pth'
```
sei-nwinski, line 114: Is there a reason why we'd have dt2 produce a model file name `model_final.pth` in no-op training, but then a `model.pt` in all other cases? Shouldn't we be consistent?

sei-nwinski: Disregard

sei-nwinski (approved):

PR #88, opened 2022-05-16
Removed pytest dependencies on workspace.

sei-amellinger:


test/test_data.py
```diff
@@ -800,41 +801,46 @@ def mock_load_tabular_data(mc):
...
-    TestCase().assertDictEqual(func_labels, test_labels)
-    assert test_source == func_source
+        test_source = "training dataset config"
+        func_labels, func_source = jb_data.get_label_mapping(train_config=train_config, show_source=True)
+        TestCase().assertDictEqual(func_labels, test_labels)
+        assert test_source == func_source
 
-    func_labels = jb_data.get_label_mapping(model_config=model_config, train_config=train_config, show_source=False)
-    TestCase().assertDictEqual(func_labels, test_labels)
+        test_source = "training dataset config"
```
sei-nwinski, line 836: I've got other changes going in so I'll take care of this, but I just wanted to confirm that this line should actually read "model config".

sei-amellinger: Hmm, let me go check why it passed for me.

sei-amellinger: If I change it to "model config" it fails.  I didn't change the actual guts of the function. I just added three lines at the top (with clause, and the two testing_utils). It is an artifact of the way the differ functions that this shows up this way.

I'll look at how/why the source thing works that way.

test/testing_utils.py
```diff
@@ -0,0 +1,205 @@
+#! /usr/bin/env python3
```
sei-nwinski, line 1: I'm not a fan of this filename, because I think it's too close to `test/test_utils.py`. Can I rename it? Maybe just `utils.py` that lives in the `test/` directory?

sei-nwinski (approved):

PR #89, opened 2022-05-17
Bugfix/jb 423 brambox logging

sei-wshaw:
Give setup_logger a "log_filter_class" parameter that takes a subclass of logging.Filter. A filter is created that filters adjacent duplicates from the log. The motivation for this is that Brambox logging is often too chatty and not useful. 

Test with test_jb_logger.py, which shows that when RemoveDuplicatesFilter is passed as a parameter to setup_logger, adjacent duplicate messages are not logged.


juneberry/jb_logging.py
```diff
@@ -58,7 +58,8 @@
 
 
 @functools.lru_cache()
-def setup_logger(log_file, log_prefix, dist_rank=0, name="juneberry", log_to_console=True, level=logging.INFO):
+def setup_logger(log_file, log_prefix, dist_rank=0, name="juneberry", log_to_console=True, level=logging.INFO,
+                 log_filter=None):
```
sei-amellinger, line 62: Since log filter is a class and not a filter instance, we probably need to add a some API comments, and/or change the name.  When I first say "ch.addFilter(log_filter())" I was confused about the "()" part.

sei-amellinger (approved):

PR #91, opened 2022-05-17
Added ability to set strict when loading pytorch models.

sei-amellinger:
This only affects training when using previous models or evaluation.

### DIFF
--- a/juneberry/config/model.py
+++ b/juneberry/config/model.py
@@ -103,6 +103,10 @@ class PytorchOptions(Prodict):
     lr_step_frequency: str
     optimizer_args: Prodict
     optimizer_fn: str
+    strict: bool
+
+    def init(self):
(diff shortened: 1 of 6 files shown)


juneberry/config/model.py
```diff
@@ -103,6 +103,10 @@ class PytorchOptions(Prodict):
     lr_step_frequency: str
     optimizer_args: Prodict
     optimizer_fn: str
+    strict: bool
+
+    def init(self):
+        self.strict = True
```
sei-nwinski, line 109: Several other PyTorchOptions also have default values, though we seem to set those elsewhere (typically in pytorch/utils.py). Should we be moving those defaults here into the config? Now, or is that future work?

sei-nwinski (approved):

PR #100, opened 2022-06-02
patch evaluation

sei-nmvanhoudnos:
This is a big one

* Includes Nick's code for experiment configs that can descend into lists
* Includes Andrew's code on variation plots 
* Has a fair number of old commits that didn't go anywhere -- we just went with another library for that work. Definitely squash.


sei-nwinski:
I've completed my current review and pushed a bunch of fixes. There are still some problems here that will need to be addressed before this can be merged. One of the tests in `test_transform_manager.py` is failing and there are some other problems that are popping up during the system test related to torchsummary.

sei-amellinger (approved):

PR #108, opened 2022-07-15
Added support for automatic creation of a workspace as a python package

sei-nmvanhoudnos:
This is a little hackish, but appears to work.


sei-nmvanhoudnos:
This is what happens when I don't test things well enough. Additional changes incoming.

sei-amellinger (approved):

PR #110, opened 2022-07-20
Jb 459 experiment with pretrained models

sei-amellinger:
I added a 'train' flag which defaults to 'true' to the models stanza in the experiment config (not outline). If setting the flag to false, then jb_experiment_to_rules will NOT general a rule for jb_train for the model and therefore no doit action is created. The train flag is set to false by jb_generate_watermark_eval so that the model will not be trained.

To test you can set the flag anywhere in the chain and the re-generate downstream configs. Let me know if you want to insert it somewhere.  Remmeber the chain is: jb_generate_watermark_eval -> experiment config -> rules.json -> main_dodo.py.

Probably the easiest practical thing to get the benefit now is to modify the experiment config and delete the rules.json and main_dodo.py JB should automatically regen those two files.

### DIFF
--- a/bin/jb_experiment_to_rules
+++ b/bin/jb_experiment_to_rules
@@ -121,7 +121,10 @@ class RuleMaker:
             # Predictions
             if model.tests:
                 for test in model.tests:
-                    inputs = [model_mgr.get_model_path(), test.dataset_path]
+                    if model.train:
+                        inputs = [model_mgr.get_model_path(), test.dataset_path]
+                    else:
+                        inputs = [test.dataset_path]
                     outputs = model_mgr.get_evaluation_output_list(test.dataset_path)
                     command = [JB_EVALUATE_COMMAND, model_mgr.model_name, test.dataset_path]
                     if test.classify:
--- a/bin/jb_experiment_to_rules
+++ b/bin/jb_experiment_to_rules
@@ -124,6 +124,9 @@ class RuleMaker:
                     if model.train:
                         inputs = [model_mgr.get_model_path(), test.dataset_path]
                     else:
+                        # If the model isn't being trained, then a model file doesn't have to exist.
+                        # The model architecture can create a fully functional pretrained model from the
+                        # underlying platform such as torchvision pretrained models.
                         inputs = [test.dataset_path]
                     outputs = model_mgr.get_evaluation_output_list(test.dataset_path)
                     command = [JB_EVALUATE_COMMAND, model_mgr.model_name, test.dataset_path]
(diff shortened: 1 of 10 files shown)


sei-asinha (approved):

sei-wshaw (commented):
From a code perspective, I'm OK with these changes. I'll leave it to Anusha and Nathan to test and decide it's performing as advertised.

juneberry/pytorch/evaluation/evaluator.py
```diff
@@ -226,9 +226,12 @@ def obtain_model(self) -> None:
...
         # Load the weights into the model.
-        logger.info(f"Loading model weights...")
-        self.model = pyt_utils.load_model(self.model_manager.get_pytorch_model_path(), self.model,
-                                          self.model_config.pytorch.strict)
+        model_path = self.model_manager.get_pytorch_model_path()
+        if model_path.exists():
+            logger.info(f"Loading model weights...")
+            self.model = pyt_utils.load_model(model_path, self.model, self.model_config.pytorch.strict)
+        else:
+            logger.warning(f"No 'model.pt' found, running with plain model. Expected to find: {model_path}")
```
sei-nwinski, line 234: What does a "plain model" look like?

sei-amellinger: Plain (unadorned) in the sense that we didn't load additional weights. So whatever the architecture component provided which could have anything.

sei-amellinger:
> From a code perspective, I'm OK with these changes. I'll leave it to Anusha and Nathan to test and decide it's performing as advertised.

Anusha approved it so I guess it is good to go. :)

bin/jb_experiment_to_rules
```diff
@@ -117,7 +121,10 @@ class RuleMaker:
             # Predictions
             if model.tests:
                 for test in model.tests:
-                    inputs = [model_mgr.get_model_path(), test.dataset_path]
+                    if model.train:
+                        inputs = [model_mgr.get_model_path(), test.dataset_path]
+                    else:
+                        inputs = [test.dataset_path]
```
sei-nwinski, line 127: So if the user downloads a model file from some external source and places it in the model directory, that shouldn't count as an input to the eval command?

sei-amellinger: In that case yes. But we can't discriminate between whether they downloaded it manually and placed it there or the model architecture downloaded it into the cache and is using it from there.  I'll add a comment to clarify that.

sei-nwinski (approved):

PR #112, opened 2022-07-20
Jb 422 Hyperparameter Tuning for PyTorch models

sei-nwinski:
This PR will incorporate hyperparameter tuning for PyTorch models in Juneberry using Ray Tune. 

PR #10 in the juneberry-example-workspace repo is meant to accompany this PR. In that PR (#10), you'll find an example tuning config and corresponding experiment config which will perform a short tuning run on the PyTorch imagenette model found in juneberry-example-workspace.

### DIFF
--- a/bin/jb_experiment_to_rules
+++ b/bin/jb_experiment_to_rules
@@ -41,6 +41,7 @@ logger = logging.getLogger(__name__)
 JB_TRAIN_COMMAND = 'jb_train'
 JB_EVALUATE_COMMAND = 'jb_evaluate'
 JB_REPORT_COMMAND = 'jb_report'
+JB_TUNE_COMMAND = 'jb_tune'
 
 
 class RuleMaker:
@@ -58,6 +59,7 @@ class RuleMaker:
     def save(self, file_path):
 
         self.add_training()
+        self.add_tuning()
         self.add_eval()
         self.add_reports()
 
@@ -100,6 +102,36 @@ class RuleMaker:
             if model.filters is not None:
                 self.add_filters(model.filters, model_mgr, model_config.training_dataset_config_path,
                                  model['train_rule_id'])
+
+    def add_tuning(self):
+        for tuning_run in self.experiment_config.tuning:
+            model_mgr = ModelManager(tuning_run.model)
+            model_config = ModelConfig.load(model_mgr.get_model_config())
+
+            # Tuning rules
+            workflow = self.builder.get_workflow("main")
+            inputs = [model_mgr.get_model_config(), model_config.training_dataset_config_path, tuning_run.tuning_config]
+            outputs = model_mgr.get_tuning_output_list()
+            command = [JB_TUNE_COMMAND, model_mgr.model_name, tuning_run.tuning_config]
+
+            doc = " ".join([str(x) for x in command])
+            clean_extras = model_mgr.get_tuning_clean_extras_list()
+            requirements = []
+
+            # Add a temporary property for now
+            tuning_run['tune_rule_id'] = workflow.add_rule(inputs, outputs, command, doc, clean_extras, requirements)
+
+            # Dry run rules
+            workflow = self.builder.get_workflow("dryrun")
+            inputs = [model_mgr.get_model_config(), model_config.training_dataset_config_path, tuning_run.tuning_config]
+            outputs = model_mgr.get_tuning_dryrun_output_list()
+            command = [JB_TUNE_COMMAND, model_mgr.model_name, tuning_run.tuning_config, '--dryrun']
+            doc = " ".join([str(x) for x in command])
+            clean_extras = model_mgr.get_tuning_dryrun_clean_extras_list()
+            requirements = []
+
+            # Add a temporary property for now
+            tuning_run['dryrun_rule_id'] = workflow.add_rule(inputs, outputs, command, doc, clean_extras, requirements)
 
     def add_eval(self):
         self.tag_map = {}
--- a/bin/jb_experiment_to_rules
+++ b/bin/jb_experiment_to_rules
@@ -104,34 +104,39 @@ class RuleMaker:
                                  model['train_rule_id'])
 
     def add_tuning(self):
-        for tuning_run in self.experiment_config.tuning:
-            model_mgr = ModelManager(tuning_run.model)
-            model_config = ModelConfig.load(model_mgr.get_model_config())
+        if self.experiment_config.tuning is not None:
+            for tuning_run in self.experiment_config.tuning:
+                model_mgr = ModelManager(tuning_run.model)
+                model_config = ModelConfig.load(model_mgr.get_model_config())
 
-            # Tuning rules
-            workflow = self.builder.get_workflow("main")
-            inputs = [model_mgr.get_model_config(), model_config.training_dataset_config_path, tuning_run.tuning_config]
-            outputs = model_mgr.get_tuning_output_list()
-            command = [JB_TUNE_COMMAND, model_mgr.model_name, tuning_run.tuning_config]
+                # Tuning rules
+                workflow = self.builder.get_workflow("main")
+                inputs = [model_mgr.get_model_config(), model_config.training_dataset_config_path,
+                          tuning_run.tuning_config]
+                outputs = model_mgr.get_tuning_output_list()
+                command = [JB_TUNE_COMMAND, model_mgr.model_name, tuning_run.tuning_config]
 
-            doc = " ".join([str(x) for x in command])
-            clean_extras = model_mgr.get_tuning_clean_extras_list()
-            requirements = []
+                doc = " ".join([str(x) for x in command])
+                clean_extras = model_mgr.get_tuning_clean_extras_list()
+                requirements = []
 
-            # Add a temporary property for now
-            tuning_run['tune_rule_id'] = workflow.add_rule(inputs, outputs, command, doc, clean_extras, requirements)
+                # Add a temporary property for now
+                tuning_run['tune_rule_id'] = workflow.add_rule(inputs, outputs, command, doc, clean_extras,
+                                                               requirements)
 
-            # Dry run rules
-            workflow = self.builder.get_workflow("dryrun")
-            inputs = [model_mgr.get_model_config(), model_config.training_dataset_config_path, tuning_run.tuning_config]
-            outputs = model_mgr.get_tuning_dryrun_output_list()
-            command = [JB_TUNE_COMMAND, model_mgr.model_name, tuning_run.tuning_config, '--dryrun']
-            doc = " ".join([str(x) for x in command])
-            clean_extras = model_mgr.get_tuning_dryrun_clean_extras_list()
-            requirements = []
+                # Dry run rules
+                workflow = self.builder.get_workflow("dryrun")
+                inputs = [model_mgr.get_model_config(), model_config.training_dataset_config_path,
+                          tuning_run.tuning_config]
+                outputs = model_mgr.get_tuning_dryrun_output_list()
+                command = [JB_TUNE_COMMAND, model_mgr.model_name, tuning_run.tuning_config, '--dryrun']
+                doc = " ".join([str(x) for x in command])
+                clean_extras = model_mgr.get_tuning_dryrun_clean_extras_list()
+                requirements = []
 
-            # Add a temporary property for now
-            tuning_run['dryrun_rule_id'] = workflow.add_rule(inputs, outputs, command, doc, clean_extras, requirements)
+                # Add a temporary property for now
+                tuning_run['dryrun_rule_id'] = workflow.add_rule(inputs, outputs, command, doc, clean_extras,
+                                                                 requirements)
 
     def add_eval(self):
         self.tag_map = {}
--- a/bin/jb_train
+++ b/bin/jb_train
@@ -27,8 +27,8 @@ import logging
 import sys
 
 import juneberry.filesystem as jbfs
-import juneberry.loader as jb_loader
 import juneberry.scripting as jbscripting
+from juneberry.training.trainer_factory import TrainerFactory
 
 logger = logging.getLogger("juneberry.jb_train")
 
@@ -55,20 +55,6 @@ def setup_args(parser) -> None:
     parser.add_argument('--resume', default=False, action='store_true',
                         help='If the trainer supports checkpointing, then this instructs the trainer to resume training'
                              'from the last checkpoint.')
-
-
-def setup_opt_args(args: argparse.Namespace):
-    """
-    Set up optional args from the command line.
-    :param args: A Namespace object built from the attributes parsed out of the command line.
-    """
-    opt_args = {}
-    opt_arg_keys = ['resume']
-    for key in opt_arg_keys:
-        if key in vars(args):
-            opt_args[key] = vars(args)[key]
-
-    return opt_args
 
 
 def main():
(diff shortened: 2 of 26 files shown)


bin/jb_experiment_to_rules
```diff
@@ -101,6 +103,41 @@ class RuleMaker:
                 self.add_filters(model.filters, model_mgr, model_config.training_dataset_config_path,
                                  model['train_rule_id'])
 
+    def add_tuning(self):
+        if self.experiment_config.tuning is not None:
+            for tuning_run in self.experiment_config.tuning:
+                model_mgr = ModelManager(tuning_run.model)
+                model_config = ModelConfig.load(model_mgr.get_model_config())
```
sei-amellinger, line 110: We should be trying to use the lab when possible.  Consider lab.load_model_config()

bin/jb_train
```diff
@@ -79,120 +66,67 @@ def main():
...
-        trainer.native = True
+    trainer.set_output_format(native=not sprout.skip_native, onnx=sprout.onnx)
 
-    if args.dryrun:
+    if sprout.dryrun:
         trainer.dry_run()
     else:
         # Allocate the GPUs the user asked for, getting the current number if None are required.
-        trainer.num_gpus = trainer.check_gpu_availability(lab.profile.num_gpus)
+        trainer.num_gpus = trainer.check_gpu_availability(trainer_factory.lab.profile.num_gpus)
```
sei-amellinger, line 122: Why use the lab out of the trainer_factory?

bin/jb_train
```diff
@@ -79,120 +66,67 @@ def main():
...
     # Set the output format for the trainer.
-    trainer.onnx = args.onnx
-    trainer.native = not args.skipNative
-
-    # If no output format was set, choose the native format by default even if
-    # the user requested it to be skipped.
-    if not (trainer.onnx or trainer.native):
-        logger.warning(f"An output format was not set. Choosing the native format.")
-        trainer.native = True
+    trainer.set_output_format(native=not sprout.skip_native, onnx=sprout.onnx)
```
sei-amellinger, line 116: Should this go into the factory?

bin/jb_train
```diff
@@ -79,120 +66,67 @@ def main():
...
         trainer.dry_run()
     else:
         # Allocate the GPUs the user asked for, getting the current number if None are required.
-        trainer.num_gpus = trainer.check_gpu_availability(lab.profile.num_gpus)
+        trainer.num_gpus = trainer.check_gpu_availability(trainer_factory.lab.profile.num_gpus)
 
-        if lab.profile.max_gpus is not None:
-            if trainer.num_gpus > lab.profile.max_gpus:
-                logger.info(f"Maximum numbers of GPUs {trainer.num_gpus} being capped to {lab.profile.max_gpus} "
+        if trainer_factory.lab.profile.max_gpus is not None:
```
sei-amellinger, line 124: Should all this gpu setting go into the factory?

bin/jb_experiment_to_rules
sei-nwinski: I was thinking about this for a bit and had some questions:
 - Labs aren't used anywhere else in `jb_experiment_to_rules`, so it feels like if I set one up, I'd have to rewrite all the places in the script where a lab.load would apply. Seems like scope creep? Maybe a separate issue to address the lack of a lab in this script?
 - This script also gets called by `jb_run_experiment`, which sets up its own lab. Would it make sense to for that same lab to also be used by `jb_experiment_to_rules`?

bin/jb_train
sei-nwinski: It felt like the right thing to do.

The trainer was constructed using the lab in the factory. What if there was some operation in the Factory that altered the GPU value in the Factory's Lab such that it is now a different value than `jb_train`'s Lab? Which value would we want to use? I thought the Factory Lab would matter more.

bin/jb_tune
```diff
@@ -0,0 +1,105 @@
...
+
+
+def setup_args(parser) -> None:
+    """
+    Adds arguments to the parser
+    :param parser: The parser in which to add arguments.
+    """
+    parser.add_argument('modelName', help='Name of the directory in the workspace "models" directory containing the '
+                                          'model \'config.json\' of the model to be tuned.')
+    parser.add_argument('tuningConfig', help='Name of the tuning config to apply to the model.')
```
sei-amellinger, line 45: Is this a path and if so relative to what?

bin/jb_train
sei-nwinski: This comment will no longer apply, since these GPU steps are being moved to the TrainerFactory.

bin/jb_train
sei-nwinski: I was able to accomplish this by adding a 'native' and 'onnx' arg to the TrainerFactory's `get_trainer` method.

bin/jb_train
sei-nwinski: Yes, it will be in the TrainerFactory.

docs/specs/tuning_configuration_specification.md
```diff
@@ -0,0 +1,153 @@
...
+Linux style version of the **format** of the file. Not the version of 
+the data, but the version of the semantics of the fields of this file. 
+The current version: 0.1.0
+
+## num_samples
+This integer indicates how many times to sample from the hyperparameter space. Each "sample" 
+corresponds to a selection of hyperparameter values from the search space. The model is then 
+trained for a "trial" using those hyperparameters.
+
+## scheduler
```
sei-amellinger, line 68: This and the search algo should probably start with the "*Optional*" tag we have for other things.

bin/jb_tune
sei-nwinski: Good question. I really couldn't decide what the proper "home" for the tuning configs should be. It didn't make sense to associate a tuning config file with a model, because you could have a situation where the same tuning config could be used for multiple models. It also didn't make sense to associate with just an experiment, since you could have multiple models associated with the experiment that happen to be "too different" for the same tuning config to apply to each of them.

For the purposes of juneberry-example-workspace, I ended up saving the example tuning config in its own separate directory named `tuning`. I think ultimately, this arg should receive the location of the target tuning config no matter where the user decides to store it. Similar to the `-f` arg in `jb_report`.

juneberry/config/experiment.py
```diff
@@ -54,9 +54,9 @@ class ModelTest(Prodict):
 
 class Model(Prodict):
     filters: List[str]
-    name: str
```
sei-amellinger, line 57: Why did you remove the model name?

docs/specs/tuning_configuration_specification.md
sei-nwinski: Good point. Added.

bin/jb_tune
sei-amellinger: So this sound like `data_sets`. We have a common location, but it isn't required because we can use it elsewhere. However, it is workspace relative.

juneberry/config/experiment.py
sei-nwinski: I thought I had a reason related to training a model that was being tuned, but I may have solved that elsewhere. I'll have to do some more testing with it restored.

juneberry/config/model.py
```diff
@@ -279,6 +279,24 @@ def to_json(self):
...
 
+    def adjust_attributes(self, adjustment_dict: dict) -> Prodict:
+        """
+        This method is responsible for replacing attributes in a ModelConfig.
+        :param adjustment_dict: A dictionary where the keys indicate which ModelConfig attributes
+        to change, and the corresponding value indicates what the attribute value should be changed to.
+        :return: A ModelConfig whose attributes have been adjusted to the desired values.
+        """
+        # Convert the current ModelConfig to a dictionary.
+        as_dict = conf_utils.prodict_to_dict(self)
```
sei-amellinger, line 290: Why do you have to convert this to a normal dict?

docs/specs/tuning_configuration_specification.md
```diff
@@ -0,0 +1,153 @@
...
+    "scheduler": {
+        "fqcn": <fully qualified name of the ray.tune scheduler class to use for scheduling>,
+        "kwargs": { <OPTIONAL> kwargs to be passed to the scheduler class during its construction }
+    },
+    "search_algorithm": {
+        "fqcn": <fully qualified name of the ray.tune search algorithm class to use for hyperparameter selection>,
+        "kwargs": { <OPTIONAL> kwargs to be passed to the search algorithm class during its construction }
+    },
+    "search_space": {
+        "model_config_attribute_name": {
```
sei-amellinger, line 30: I recommend changing this pattern to an array of elements of "name", "fcqn", "kwargs."  Pros allows use of schema validation, cons a person can add duplicate versions, but in general I find that using values as the keys ends up being more work.

juneberry/schemas/tuning_schema.json
```diff
@@ -0,0 +1,41 @@
...
+    },
+    "type": "object",
+    "description": "A configuration for tuning hyperparameters in a Juneberry model.",
+    "properties": {
+        "description": { "type": "string" },
+        "format_version": { "type": "string" },
+        "num_samples": { "type": "integer" },
+        "scheduler": { "$ref": "#/definitions/plugin_item" },
+        "search_algorithm": { "$ref": "#/definitions/plugin_item" },
+        "search_space": { "type":  "object" },
```
sei-amellinger, line 22: Should search space be required?

juneberry/config/experiment.py
sei-nwinski: I'll be adding it back.

juneberry/config/model.py
sei-nwinski: You're right, I can stick with the Prodict.

juneberry/schemas/tuning_schema.json
sei-nwinski: Yeah, you're right. I'll add it to the schema and remove the else statement in the Tuner that will exit if there's no search space.

docs/specs/tuning_configuration_specification.md
sei-nwinski: Sounds reasonable. I'll get it implemented.

docs/specs/tuning_configuration_specification.md
```diff
@@ -0,0 +1,168 @@
...
+"model_architecture.args.num_classes". 
+
+### kwargs
+This stanza contains all the arguments to pass in to the sampling function during its construction.
+
+## timestamp
+**Optional:** Timestamp (ISO format) indicating when the config was last modified.
+
+## trial_resources
+The trial resources are used to indicate what machine resources should be allocated per 'trial'. A
```
sei-amellinger, line 130: What would it mean to have 2 cpus and 2 gpus?  Isn't this an either or?

docs/specs/tuning_configuration_specification.md
```diff
@@ -0,0 +1,168 @@
...
+This integer indicates how many GPUs should be allocated to the trial.
+
+## tuning_parameters
+Tuning parameters control what aspects the Tuner will attempt to optimize during model training and 
+how it should perform that optimization. By default, the Tuner will attempt to minimize the last 
+reported training loss value.
+
+### checkpoint_interval
+This integer indicates how often to save model checkpoints during tuning. For example, when this field 
+is set to 1, a model checkpoint will be saved every time a round of metrics is sent to the tuner. If
```
sei-amellinger, line 147: When you say "round" do you mean "sample"?

docs/specs/tuning_configuration_specification.md
```diff
@@ -0,0 +1,168 @@
...
+
+### checkpoint_interval
+This integer indicates how often to save model checkpoints during tuning. For example, when this field 
+is set to 1, a model checkpoint will be saved every time a round of metrics is sent to the tuner. If 
+this field is set to 0, then no model checkpoint directories will ever be saved. 
+
+**NOTE:** Some tuning features, such as certain types of schedulers, may rely on checkpointing in order 
+to work properly.  
+
+### metric
```
sei-amellinger, line 153: Does this options come from the "search_algorithm"?  How do I know which are available?

juneberry/config/experiment.py
```diff
@@ -79,6 +80,11 @@ class Filter(Prodict):
     outputs: List[str]
 
 
+class Tuning(Prodict):
```
sei-amellinger, line 83: Should this an other tuning changes go into the 'experiment_configuration_specification.md'`?

docs/specs/tuning_configuration_specification.md
```diff
@@ -0,0 +1,168 @@
...
+a quick, high-level explanation of several tuning components that can be controlled using the tuning 
+config. Of the Ray Tune components described in this link, the tuning config allows you to define 
+which search space, search algorithm, and scheduler to use during your tuning run.
+
+# Schema
+
+```
+{
+    "description": <OPTIONAL text description of the tuning performed by this file>,
+    "format_version": <Linux style version string of the format of this file>,
```
sei-amellinger, line 21: Missing file_path

docs/specs/tuning_configuration_specification.md
sei-nwinski: I don't think it's an either/or. My understanding was CPU affects how many trials run concurrently, while GPUs affects CUDA_VISIBLE_DEVICES.

A resource: https://docs.ray.io/en/latest/tune/tutorials/tune-resources.html

juneberry/config/tuning.py
```diff
@@ -0,0 +1,132 @@
...
+        self.metric = 'loss'
+        self.mode = 'min'
+        self.scope = 'last'
+
+
+class TuningConfig(Prodict):
+    FORMAT_VERSION = '0.1.0'
+    SCHEMA_NAME = 'tuning_schema.json'
+    description: str
+    file_path: Path
```
sei-amellinger, line 65: missing format_version

docs/specs/tuning_configuration_specification.md
sei-nwinski: I don't think so. Since we have a parameter "num_samples" in the tuning config, sample to me means how many times we make a selection of parameters out of the search space. The model config then gets modified to match all of the hyperparameter choices in the sample, and then training starts. 

Since that model may train for multiple epochs, and after each epoch the metrics get sent to the tuner to decide whether or not to terminate training early, I was referring to the potential multiple exchanges of metrics as rounds.

juneberry/filesystem.py
```diff
@@ -751,6 +755,14 @@ def get_train_scratch_path(self) -> Path:
         """ :return: Path to a directory for 'scratch' outputs from training. """
         return self.get_train_root_dir() / "scratch"
 
+    def get_tuning_dir(self) -> Path:
+        """ :return: Path to a directory containing files related to hyperparameter tuning. """
+        return self.get_train_root_dir() / "tuning"
+
+    def get_tuning_log(self) -> Path:
```
sei-amellinger, line 762: How does this work? When I am reading through the code it looks like the path should be:
`models/<model>/train/tuning/log.txt`
But looking at my code I have a file at:
`models/imagenette_160x160_rgb_unit_test_pyt_resnet18/train/tuning/_tuning_attempt_2022-07-26_09-36-59/log.txt`
How does get tuning log work?

docs/specs/tuning_configuration_specification.md
sei-nwinski: I'm assuming you're referring to "metric" here.

The search_algorithm is responsible for deciding how to select hyperparameter values out of the hyperparameter search space, to be substituted in to the model config prior to training. My understanding is it's not related to the metric at all.

After the hyperparameters have been selected and the new model config gets created to represent those adjustments, model training happens. During that, the training metrics get reported to the Tuner on a per epoch basis. 

So you need to define how you want the Tuner to optimize those reported metrics. You do that with the combination of metric/mode/scope. Metric indicates which of the training metrics you want the Tuner to examine, mode indicates if you want to minimize or maximize the value of the metric (e.g. you'd want to min loss or max accuracy), and then scope controls how many "reports" of the metric value you want the Tuner to consider (e.g. just the last value, and average of the last 5, etc).

juneberry/config/experiment.py
sei-nwinski: Yes. I caught that while working on experiment unit tests yesterday and already have the changes drafted but not yet pushed.

docs/specs/tuning_configuration_specification.md
sei-nwinski: It shouldn't exist; my config is wrong. Bad copy/paste?

juneberry/config/tuning.py
sei-nwinski: Will change file_path -> format_version.

sei-nwinski: Wait, no. It's at the top.

juneberry/tuning/tuner.py
```diff
@@ -0,0 +1,342 @@
...
+        """
+        # Retrieve the desired scheduler from the tuning config.
+        scheduler_dict = self.tuning_config.scheduler
+        logger.info(f"Constructing tuning scheduler using fqcn: {scheduler_dict.fqcn}")
+
+        # All schedulers in Ray Tune use 'self.metric' and 'self.mode' to make decisions about
+        # terminating bad trials, altering parameters in a running trial, etc. However,
+        # tune.run() receives both of these properties directly and will complain when they are
+        # also provided to the scheduler.
+        if "metric" in scheduler_dict.kwargs:
```
sei-amellinger, line 127: I wonder if we should just die.  What if they don't match and the "right" one is in the scheduler_dict?

juneberry/filesystem.py
sei-nwinski: If you were to interrupt tuning while it's running, you'd find the log at the indicated location.

If you look inside `juneberry/tuning/tuner.py`, specifically the last couple lines of the file, you'll see that inside `finish_tuning` I take steps to move the log file from the "vanilla" tuning directory into the results directory that was generated for that tuning run. I thought that location would make it more clear which log file was associated with which tuning run.

Otherwise you'd potentially have a stack of log files next to a stack of tuning_attempt directories and you'd have to figure out which ones belonged together.

juneberry/tuning/tuner.py
sei-nwinski: I believe this was included to bypass a warning that Ray Tune threw. Something to the effect of "Hey you already told me the metric directly, why did you also define one in the scheduler?". I think it then went on to ignore the one in the scheduler. 

So I figured if it's being ignored anyway, just pop the value from the scheduler, and warn the user that they should update their config.

juneberry/filesystem.py
```diff
@@ -232,6 +232,17 @@ def load_file(path: str):
             sys.exit(-1)
 
 
+def load_json_lines(path: str) -> list:
```
sei-amellinger, line 235: In general our load file also supports hjson too. I think if you switch this to hjson.loads() is should be able to load vanilla json and hjson. 
But if this is really only for ".json" files then you might want to add an extension ject.

sei-amellinger (approved):

PR #120, opened 2022-08-11
JB-222-280 remove platform and model versions concepts.

sei-amellinger:
Removed notions of platform from filesystem, naming etc. This includes adding a new "platforms definitons" notion and class to be provided to routines (such as model path) that need it.

Removed the old "model versions" stuff.


juneberry/loader.py
```diff
@@ -94,6 +94,41 @@ def extract_kwarg_names(func):
...
+    class_data = fully_qualified_name.split(".")
+    path_part = ".".join(class_data[:-1])
+    name_part = class_data[-1]
+
+    return path_part, name_part
+
+
+def load_class(fqcn):
+    """
+    Loads the class from the FQCN.  The module and class are expected to exist and if not, then we exit.
```
sei-nwinski, line 112: Looking at the usages of `load_class`, I'm not sure "then we exit" ever really happens. I was going to rewrite this to reflect that we're returning False, but I have some concerns about that.

Don't we run the risk of problems down the line by returning False? It feels like the usages assume we're always getting a class back and not doing any checks for "False".

sei-amellinger: Yeah, we should be raising an error or exiting.

sei-nwinski (approved):

PR #131, opened 2022-09-01
JB-481 general cleanup

sei-nwinski:
There are a lot of files that changed, but the majority of changes involve updates to import statements to improve consistency across all files, updated doc markings, and the relocation of some files to other directories (e.g. trainer.py was moved to juneberry/training).


juneberry/architectures/pytorch/torchvision.py
```diff
@@ -33,12 +33,12 @@ class Wrapper:
     Basic wrapper for torchvision models classes
     """
 
-    def __call__(self, className, classArgs, num_classes):
+    def __call__(self, class_name, class_args, num_classes):
```
sei-amellinger, line 36: This class here was to deal as a shim with deprecated code. I'm not sure we want to change kwargs signature at this time.

juneberry/detectron2/transforms.py
```diff
@@ -54,19 +33,24 @@ class DT2NoOp:
     """
     Sample no-operation transformer with ALL operations for copy-pasting
     """
-    def apply_image(self, img: np.ndarray):
+    @staticmethod
```
sei-amellinger, line 36: Just because these can be static methods, let's not do that in case someone does indeed want to subclass this.

juneberry/helpers/data_analysis.py
```diff
@@ -255,15 +228,16 @@ def __init__(self, dataset_path_manager):
...
         self.dataset_path_manager = dataset_path_manager
         self.image_df = None
         self.anno_df = None
         self._build_image_df()
         self._build_anno_df()
 
-    def _load_json(self, path):
-        return jbfs.load_file(path)
+    @staticmethod
+    def _load_json(path):
```
sei-amellinger, line 239: I dont think this should be static.

juneberry/helpers/data_analysis.py
```diff
@@ -119,116 +100,108 @@ def __init__(self, lab, ws_key='default', dr_key='default', dataset='', train_pa
         if not self.test_anno_path:
             self.anno_sources.remove('test')
 
-    
-    def _fix_file_paths(self, paths):
+    @staticmethod
```
sei-amellinger, line 103: Probably should not be static.

juneberry/helpers/data_analysis.py
```diff
@@ -440,12 +412,14 @@ def _initialize_figure_subplots(self, fig_key, nrows, ncols, fig_key_index=None)
                 axes[r, c] = ax
         return axes
 
-    def _add_line_to_plot(self, ax, x_pos, text):
+    @staticmethod
```
sei-amellinger, line 415: Static?

sei-nwinski: I think I reverted all the staticmethod decorators that I introduced in this PR.

sei-amellinger (approved):

PR #135, opened 2022-09-08
JB-481: Addressing some TODOs

sei-nwinski:
Almost all of the changes in this branch address lingering TODOs at various locations in the Juneberry code. For the most part, the changes in a particular file will be located around a TODO comment. However, there are a few cases where the changes needed to address the TODO involve code in another file. So if it seems like a file changed arbitrarily, there's a good chance the corresponding TODO is located elsewhere.


bin/jb_process_dataset
```diff
@@ -301,8 +301,8 @@ class TFDSProcessor(DatasetProcessor):
 
         # TODO: This should ALL go into the tensorflow data module.
         # Walk through the loader batches, sampling one from each.
-        idx = 0
         for ds_idx, ds in enumerate(self.datasets):
+            idx = 0
```
sei-amellinger, line 305: Why move this?

juneberry/evaluation/evaluator.py
```diff
@@ -291,13 +291,15 @@ def perform_evaluation(self) -> None:
...
-    def perform_additional_eval(self, log_file, dataset: DatasetConfig, eval_dir_mgr: EvalDirMgr) -> None:
+    def perform_additional_eval(self, log_file, dataset: DatasetConfig, eval_dir_mgr: EvalDirMgr,
+                                dryrun: bool = False) -> None:
         """
         Allows the evaluation of a different dataset that has the same basic characteristics
         as the original data set, in terms of model layer sizes, etc.
         :param log_file: The file to log to
         :param dataset: The dataset to evaluate
         :param eval_dir_mgr: Where to place the evaluation
+        :param dryrun: Boolean indicating whether or not to conduct a dryrun of ethe eval.
```
sei-amellinger, line 302: ethe -> the

bin/jb_process_dataset
sei-nwinski: This was a bug I noticed when I was helping Jasmine troubleshoot. When you have multiple datasets here, you want to reset the index back to 0 at the start of a new index, otherwise the log messages look funny.

In the previous location you wind up logging something like:
Dataset 1
0/1600
160/1600
320/1600
and so on
Dataset 2
1600/400
1640/400
1680/400
and so on

setup.py
```diff
@@ -63,7 +63,7 @@
 
 bin_scripts = [
     'bin/jb_attack_to_rules',
-    'bin/jb_clean_predictions',
+    'bin/jb_clean_experiment_evals',
```
sei-amellinger, line 66: Can we add jb_process_dataset?

bin/jb_process_dataset
sei-amellinger: Okay. So for logging purposes only? The file names ended up being having unique numbers. If we want to clarify logging, maybe a separate tracker?

juneberry/evaluation/evaluator.py
sei-nwinski: Fixed.

setup.py
sei-nwinski: Done.

juneberry/config/training_output.py
```diff
@@ -158,11 +158,23 @@ def set_from_model_config(self, model_name: str, model_config: ModelConfig) -> N
...
         self.output.options.epochs = model_config.epochs
-        self.output.options.model_architecture = model_config.model_architecture
+
+        # Recording the model_architecture is a little more involved. If the model_architecture contains
+        # kwargs, there's a chance a "labels" kwarg could be introduced (particularly in TensorFlow trainers)
+        # which may have integer keys. Integer keys would cause problems for Prodict and JSON, so it's best
+        # to eliminate the "labels" kwarg if one is present. Historically, labels weren't included in the
+        # Juneberry's TrainingOutput anyway.
+        if model_config.model_architecture.kwargs is not None:
+            kwarg_dict = {k: v for k, v in model_config.model_architecture.kwargs.items() if k != "labels"}
```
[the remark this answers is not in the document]
sei-nwinski: I had some issues implementing the solution we discussed, but I think I found an alternative that still addresses your initial concern and is probably "cleaner" than what we were looking to do.

juneberry/config/training_output.py
```diff
@@ -158,11 +158,24 @@ def set_from_model_config(self, model_name: str, model_config: ModelConfig) -> N
         :param model_config: The model config object.
         :return: None
         """
-        self.output.results.model_name = model_name
 
         self.output.options.batch_size = model_config.batch_size
         self.output.options.epochs = model_config.epochs
+
+        # Recording the model_architecture is a little more involved. If the model_architecture contains
```
sei-amellinger, line 165: Is the intent to eliminate the labels in the output totally, or just be able to copy it? This code looks to temporarily pullit out, apply the assign then put it back. Since the assign is by ref (not copy) then the labels are in "both" (because they are the same) self.output.options.model_architecture AND model_config.model_architecture. So is it eliminate or a workaround to copy integer labels?

setup.py
sei-amellinger: Thanks!

sei-amellinger (approved):

PR #136, opened 2022-09-13
JB-489 wire no-paging up to profile

sei-amellinger:
- Fixed no paging support 
- Added check for profile name

### DIFF
--- a/bin/jb_train
+++ b/bin/jb_train
@@ -103,8 +103,6 @@ def main():
     # Now load the dataset
     dataset_config = lab.load_dataset_config(model_config.training_dataset_config_path)
 
-    # TODO: Reimplement no_paging via lab profile
-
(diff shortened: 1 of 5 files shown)


sei-wshaw (commented):
I understand what you're doing and it looks ok. What's the best way to test it?

sei-amellinger:
You can add a workspace config file with the no paging switch.  For example, this workspace config (config.json in workspace)

```
{
    "profiles": [
        {
            "name": "nopaging",
            "model": "default",
            "profile": {
                "no_paging": true
            }
        }
    ]
}
```

Can be activated during jb_train with with the profile switch `-p nopagin` and you should see that config show up in the log. I wouldn't actually let it complete as it will take up lots of memory.

sei-wshaw (approved):

PR #138, opened 2022-09-16
Fix ray version to 1.13.0 so unit tests pass

sei-wshaw:


sei-amellinger:
Should all the version numbers be bumped to 12.2 for consistency?

PR #141, opened 2022-09-26
JB-444 wire in model zoo

sei-amellinger:
Wire zoo in jb_evaluate and classification training.

To test there are currently two models on artifactory that are copies of the PT and TF2 imagenette models.  So:

1) Set JUNEBERRY_MODEL_ZOO - "export JUNEBERRY_MODEL_ZOO=https://artifactory.sei.cmu.edu:443/artifactory/etc-docker/juneberry/models/"
2) Use zoo-pt-example - "jb_evaluate zoo-pt-example data_sets/data_sets/imagenette_unit_test.json"
3) Use zoo-tf-example - "jb_evaluate zoo-tf-example data_sets/data_sets/imagenette_unit_test.json"

They should download, instantiate and evaluate normally.

### DIFF
--- a/docs/zoo.md
+++ b/docs/zoo.md
@@ -1,0 +1,37 @@
+Workspace and Experiment Overview
+==========
+
+# Introduction
+
+Juneberry supports the idea of a model zoo where config files and pre-trained models can be downloaded.
+Model zoo files are stored on remote servers in a hierarchy similar to how the models directory is
+constructed. Given a path to a zoo file of:
+
+"https://junebery.com/models/my-model/resnet.zip"
+
+- https://juneberry.com/models - The base url to where models are stored on the server.
+- my-model/resnet18.zip - A zip containing the model for the model "my-model/resnet18"
+
+# Packaging a model
+
+The contents of the zip file are the important contents of the model that is to be shared. This is usually
+
+- config.json
+- model.pt or model.h5
+- (optional) hashes.json
+
+If desired the hashes.json can be provided indicating what specific model architecture was used to generate
+this model. If this file exists and the model_architecture hash embdedded insides does NOT match the hash of 
+locally constructred model architecture summary, then the model will not be loaded and an error is generated.
+During training the "hashes-latest.json" file is generted which contains the model_architecture hash that was
+used to train the model
+
+A convenience tool is provided that will package up the zip file. To invoke specify the model and a directory
+that represents a staging are for the zoo where the zip files is to be run. The tool is expected to be run from
+the roots of the worksapce.  For exampe
+
+`python -m juneberry.zoo my-model/resnet18 ./zoo-staging`
+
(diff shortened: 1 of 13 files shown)


juneberry/zoo.py
```diff
@@ -225,9 +263,38 @@ def prepare_model_for_zoo(model_name: str, staging_zoo_dir: str, onnx: bool = Tr
...
+        else:
+            # If we have a latest hashes, then use it.
+            latest_hashes_path = model_mgr.get_latest_hashes_config()
+            if latest_hashes_path.exists():
+                zip_file.write(latest_hashes_path, model_mgr.get_hashes_config_filename())
+
     return str(model_zip_path.absolute())
 
 
+def update_hashes_file(hashes_path, model_architecture_hash: str = None):
```
sei-nwinski, line 279: There are some inconsistencies with the type for `hashes_path` here. It seems like the function treats it like a Path, but I think hashes.save() expects the arg to be a string.

sei-amellinger: Okay, I updated the signature.  In general the path/str hygiene should be updated better. At one point we started to return more strings and then wrap in Paths internally, but I don't think that has been deployed as we still return a lot of Paths. In the long term I'd like to come up with a better plan.

sei-nwinski: Agree. You can definitely see it skimming through filesystem.py (sometimes str, sometimes Path). I really liked the bit in this branch where you defined the filename as a str in one method and used it in a different one to assemble a Path. Maybe something like methods defining filenames are "private" to filesystem.py (so `_get_hashes_config_filename`) and then use a collection of those methods to create other methods that assemble and return Paths? Then if you *really* need a str somewhere, you're responsible for converting the Path?

sei-amellinger: That might be an approach. For the most part we don't really need to know the name, just the path. The zoo code wanted to add it to a zip, so it actually needed the name. But really there should probably something in the filesystem that returns a map "zippable" contents of name:path pairs so we just walk that. In my mind the filesystem is supposed to be as self contained and minimal as possible to try to prevent people from directly getting to filenames and structure. In retrospect I violated that with the zoo.

if you want (and you have time) I could switch that over.

docs/zoo.md
```diff
@@ -0,0 +1,38 @@
...
+
+- config.json
+- model.pt or model.h5
+- (optional) hashes.json
+
+When provided, the hashes.json file can confirm which model architecture was used to generate
+the model. If the model_architecture hash embedded inside the hashes.json does NOT match the hash of 
+locally constructed model architecture summary, then the model will not be loaded from the zoo and an 
+error is generated. During training, a "hashes-latest.json" file will be produced which contains the 
+model_architecture hash that was used to train the model.
```
sei-nwinski, line 28: It feels like there should be one more sentence at the end here.

sei-nwinski (approved):

PR #142, opened 2022-09-27
JB-481 TODOs part 2

sei-nwinski:
In this branch I cleaned up some more TODOs and refactored some things in the unit tests. I also moved a couple of things into the test utils so instead of test files importing other test files, they can just import the utils.

### DIFF
--- a/juneberry/config/model.py
+++ b/juneberry/config/model.py
@@ -158,7 +158,7 @@ class ModelConfig(Prodict):
     file_path: Path
     format_version: str
     lab_profile: LabProfile
-    label_mapping: typing.Union[Prodict, str]
+    label_mapping: object
     mmdetection: MMDetection
     model_architecture: ModelArchitecture
     model_transforms: List[Plugin]
--- a/juneberry/config/util.py
+++ b/juneberry/config/util.py
@@ -72,12 +72,11 @@ def require_tags(label, data: dict, tags: list) -> int:
 
 def validate_schema(data, schema_name, die_on_error=False):
     # Load the schema.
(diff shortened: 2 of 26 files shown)


juneberry/config/model.py
```diff
@@ -158,7 +158,7 @@ class ModelConfig(Prodict):
     file_path: Path
     format_version: str
     lab_profile: LabProfile
-    label_mapping: typing.Union[Prodict, str]
+    label_mapping: object
```
sei-amellinger, line 161: Why object instead of Any?

sei-nwinski: I guess my first reaction doesn't go to Any because it feels too...permissive? I'm not firmly attached to what gets used, so long as it works. Prodict was not happy with the Union.

juneberry/config/util.py
```diff
@@ -72,12 +72,11 @@ def require_tags(label, data: dict, tags: list) -> int:
 
 def validate_schema(data, schema_name, die_on_error=False):
     # Load the schema.
-    # TODO: Should this be routed through the jb_fs load chokepoint?
-    schema = hjson.loads(pkgutil.get_data('juneberry', f"schemas/{schema_name}"))
+    schema_filename = pkg_resources.resource_filename('juneberry', f"schemas/{schema_name}")
```
sei-amellinger, line 75: Did you want to make any other changes here (e.g. importlib?) or is this good to go?
BTW: We have some unused imports.

sei-nwinski: After discussion, we decided to revert the previous code and update the TODO.

sei-amellinger (approved):

PR #144, opened 2022-09-30
JB-239 Added detections gzip support for prediction files

sei-jwidjaja:
After running the jb_evaluate, we should see detections_anno.json.gz instead of detections_anno.json and when passed into the jb_report command, it should be able to read in the gzip file.

### DIFF
--- a/juneberry/config/coco_utils.py
+++ b/juneberry/config/coco_utils.py
@@ -30,13 +30,16 @@ from pathlib import Path
 from random import shuffle as rand_shuffle
 import sys
 from typing import Dict, List
+import gzip
+import json
+import os
 
 import numpy as np
 from PIL import Image, ImageDraw
 
 from juneberry.config.dataset import DatasetConfig
 from juneberry.config.coco_anno import CocoAnnotations
-import juneberry.filesystem as jb_fs
+import juneberry.filesystem as jbfs
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


juneberry/config/coco_utils.py
```diff
@@ -30,13 +30,16 @@
 from random import shuffle as rand_shuffle
 import sys
 from typing import Dict, List
+import gzip
```
sei-amellinger, line 33: Cleanup ordering.

juneberry/config/coco_utils.py
```diff
@@ -30,13 +30,16 @@
...
+import json
+import os
 
 import numpy as np
 from PIL import Image, ImageDraw
 
 from juneberry.config.dataset import DatasetConfig
 from juneberry.config.coco_anno import CocoAnnotations
```
sei-nwinski, line 41: Can you also swap the ordering so the coco_anno import is above the dataset import?

juneberry/config/coco_utils.py
```diff
@@ -30,13 +30,16 @@
...
-import juneberry.filesystem as jb_fs
+import juneberry.filesystem as jbfs
```
sei-nwinski, line 42: Actually we want this to stay as `jb_fs` so it's consistent across all files. It was a recent "cleanup" change.

juneberry/config/coco_utils.py
```diff
@@ -357,11 +379,29 @@ def save_predictions_as_anno(data_root: Path, dataset_config: str, predict_file:
...
+        logger.info(f"Large file detected: images={num_images} and categories={num_categories}")
+
+        remove_indents_from_json(output_file)
+
+        with gzip.open(str(output_file) + '.gz', 'wb') as f_gzip, open(str(output_file)) as f_in:
+             predict_data = f_in.read()
+             bin_predict_data = bytearray(predict_data, encoding='utf-8')
+             f_gzip.write(bin_predict_data)
+
+        os.remove(str(output_file))
```
sei-nwinski, line 402: Another opportunity to just unlink your Path.

PR #145, opened 2022-09-30
Pytorch and tensorflow metrics plugin integration

sei-wshaw:


juneberry/pytorch/evaluation/evaluator.py
```diff
@@ -32,7 +32,7 @@
 import juneberry.config.dataset as jb_dataset
 from juneberry.config.dataset import DatasetConfig
 from juneberry.config.eval_output import EvaluationOutput
-from juneberry.config.model import ModelConfig
+from juneberry.config.model import ModelConfig, Plugin
```
sei-nwinski, line 35: Any reason why you chose to import Plugin this way instead of from `juneberry.config.plugin`?

sei-wshaw: No reason, other than I've done it that way before. Is it better to switch it?

sei-nwinski: I just assumed it was better to import directly from the source rather than from a file that imports from the source. I can fix it where I notice it. There were a few other import ordering things I was going to adjust.

juneberry/pytorch/trainer.py
```diff
@@ -194,14 +203,24 @@ def setup(self):
...
 
-        self.history = {'loss': [], 'accuracy': [], 'val_loss': [], 'val_accuracy': [], 'epoch_duration': [], 'lr': []}
+        self.history = {'loss': [], 'val_loss': []}
+
+        # Create an entry in the history for each metrics plugin in the config.
+        # TODO what if one of our metrics is called "loss?" That will step on the already
+        #  existing "loss" entry in self.history.
+        #  For now, log an error if "loss" is specified in the training_metrics section.
+        for plugin in self.metrics_plugins:
+            self.history[plugin["kwargs"]["name"]] = []
```
sei-nwinski, line 219: So a plugin would require a `name` kwarg in order for this to work?

sei-wshaw: Yes. So the general idea was to be able to switch out one metric for another, using the same name. For example, you might have a metric with the name "accuracy" which is initially computed using sklearn for example. Later you might want to calculate "accuracy" using a torchmetrics plugin. The name is kind of the handle for the metric.

sei-nwinski: Maybe we need to enforce that via schema then. Right now Plugin just needs an `fqcn` and `kwargs` to be valid. A metrics plugin has that extra requirement.

sei-wshaw: So something like metrics_plugin.json in the schemas

juneberry/pytorch/trainer.py
```diff
@@ -225,14 +247,21 @@ def start_epoch_phase(self, train: bool):
...
+            for plugin in self.metrics_plugins:
+                self.training_metrics_lists[plugin.kwargs["name"]] = \
+                    [torch.zeros(1, dtype=torch.float64).cuda() for i in range(self.num_gpus)]
+
+        result["loss_list"] = []
 
         # Start off with empty metrics
-        return {'losses': [], 'accuracies': []}
+        for plugin in self.metrics_plugins:
+            result[plugin.kwargs["name"] + "_list"] = []
```
sei-amellinger, line 262: Just to be a little more clear, can we move the "_list" into a constant such a "LIST_SUFFIX" and append it when needed? Also when removing it we can instead do key.rstrip(LIST_SUFFIX)?

sei-amellinger (approved):

PR #152, opened 2022-11-15
Add force cuda flag to dt2 for building on github.

sei-amellinger:


sei-nwinski:
During testing, I used the Dockerfile to build a fresh cudadev on a non-GPU machine. I then tested Juneberry in a container built from that image and encountered no problems when running any of our system tests.

PR #158, opened 2023-04-13
Fixed torchmetrics and scikit-learn ref.

sei-amellinger:
The docker files changed sklearn to scikit learn and so juneberry setup needs to reflect this. Also, torchmetrics should be part of the torch extras as they require torch.

Testing this should simply be to do a `pip install -e .` and see that sklearn (pre whatever) stuff is NOT installed.


sei-hmoore (approved):
Looks good

